PR #3, opened 2019-01-15
Getters for gas consumption

oischinger:
Gas heaters are providing some statistics under "heating.gas.consumption.heating".
This change makes the data available either as an array of values (e.g. last 8 days, last 12 months, etc) or as a single value for today, this week, this month or this year.

### DIFF
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -416,50 +416,99 @@ class ViCareSession:
         except KeyError:
             return "error"
 
-    def getGasConsumptionDay(self):
+    def getGasConsumptionHeatingDays(self):
         try:
             return self.getProperty('heating.gas.consumption.heating')['properties']['day']['value']
         except KeyError:
             return "error"
 
-    def getGasConsumptionToday(self):
+    def getGasConsumptionHeatingToday(self):
         try:
             return self.getProperty('heating.gas.consumption.heating')['properties']['day']['value'][0]
         except KeyError:
             return "error"
 
-    def getGasConsumptionWeek(self):
+    def getGasConsumptionHeatingWeeks(self):
         try:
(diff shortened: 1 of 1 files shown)


somm15:
Do you mind if I change 
getGasConsumptionDay() to getGasConsumptionDays() or getGasConsumptionPerDay()
I would rather use the name getGasConsumptionDay for getGasConsumptionDay(day)
It would be more coherent for getPrograms() for instance.

somm15:
And by the way, thanks!
It looks like my Vitodens 222W does not support this feature, so I couldn't test this API endpoint.

somm15 (changes requested):
Could you please change naming when it return multiple values ?
e.g.  def getGasConsumptionMonth(self) -->  def getGasConsumptionMonths(self)

oischinger:
Actually I found out there is also a consumption for dhw. Let me add this and change the API as you requested. 
Thanks for this nice lib by the way.

somm15 (approved):

oischinger:
Hi @somm15 
I added the getters for DHW and renamed the existing ones to have "Heating" in the name.
Also getGasConsumptionDay() is now getGasConsumptionDays() and so on.

Unfortunately this leads to a lot of functions. Please let me know if you prefer a different API. I'm not a python expert, so maybe this code is not really python style

oischinger:
oh, that was fast ;)
Thanks

oischinger:
Ups actually too fast. I just noticed that it should be "heating.gas.consumption.dhw" instead of "heating.gas.consumption.heating" in the dhw-Methods.
Sorry, copy paste error!

somm15:
Do you want to make another commit?
(that indeed a lot of function, we could split in 3 objects
- Viessmann
|--> heating
|--> hot water

PR #20, opened 2019-11-15
add new property for supply and return temperature of primary circuit

michelde:
The heatpumps also measure the supply and return temperature of the primary circuit. Would like to have this properties later on also in home assistant.

### DIFF
--- a/PyViCare/PyViCareHeatPump.py
+++ b/PyViCare/PyViCareHeatPump.py
@@ -28,3 +28,15 @@ class HeatPump(Device):
             return self.service.getProperty("heating.compressor.statistics")["properties"]["hours"]["value"]
         except KeyError:
             return "error"
+
+    def getSupplyTemperaturePrimaryCircuit(self):
+        try:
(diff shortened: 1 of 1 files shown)


crazyfx1:
What's the difference between your new property and the getSupplyTemperature property of the generic PyViCareDevice when the self.service.circuit is set to 0? That should also be the primary circuit.

michelde:
Using a heatpump we must distinguish between the circuit for the heating (e.g. water temperature which is flowing through the floor .)  and the circuit which is going to the ground / depth (having the soil). Here I want to read the supply and return temperature. 
Doing a short test shows that the values are different:
```
import sys
import logging
from PyViCare.PyViCareDevice import Device
from PyViCare.PyViCareHeatPump import HeatPump # heat pump
from PyViCare.PyViCareService import ViCareService

device=Device("email@domain","secret","token.save", "0")
heat=HeatPump("email@domain","secret","token.save", "0")
print("Device Supply Temp. ", device.getSupplyTemperature())
print("HeatPump Supply Temp ", heat.getSupplyTemperaturePrimaryCircuit())
```
returns:
```
Device Supply Temp.  24.4
HeatPump Supply Temp  17.7
```

PR #25, opened 2020-02-28
Update PyViCareGazBoiler.py

stefanwi87:
ViCare delivers information about the burner starts and burner hours (time how long the burner was used, e.g. 6200h). Added these two statistic values of burner.

### DIFF
--- a/PyViCare/PyViCareGazBoiler.py
+++ b/PyViCare/PyViCareGazBoiler.py
@@ -122,3 +122,15 @@ class GazBoiler(Device):
             return self.service.getProperty('heating.burner.current.power')['properties']['value']['value']
         except KeyError:
             return "error"
+    
+    def getBurnerHours(self):
+        try:
+            return self.service.getProperty('heating.burner.statistics')['properties']['hours']['value']
+        except KeyError:
+            return "error"
+
+    def getBurnerStarts(self):
+        try:
+            return self.service.getProperty('heating.burner.statistics')['properties']['starts']['value']
+        except KeyError:
+            return "error"


ddokupil:
Awesome, do we have maybe some more "low hanging fruits" do add easily?

oischinger:
@somm15 Could you please release a new version on pypi? 
There is already an HA Core PR to make use of those properties: https://github.com/home-assistant/core/pull/32304

Thanks alot!

vandenberghev:
I wouldn't bother; I just got a big "fuck you" e-mail from Viessmann saying they'll be closing off API access for all third-party applications.

> an evaluation has shown that you have used the web interface to Viessmann IoT Services very frequently in recent days.
>   
> On the one hand, we are delighted to know you are taking a closer look at our products and solutions. Especially since you have (apparently) found a solution for your specific use case without a description or our support from our side.
> On the other hand, it challenges us to check and channel the method and frequency of requests to our IoT Services in order to keep those stable and available for all our users.
>   
> That’s why we have now taken the decision to limit access to our API. From the coming week on, a threshold will become active that prevents from unauthorized use of the API. This will affect all third-party solutions, which can thereby no longer be used.
>   
> We are aware that the demand for APIs to our technologies for integration into other solutions and third-party systems is increasing. To provide you with an opportunity for this, we will launch a portal in the next two months for all developers to get access to our API. With the Viessmann Developer Portal, we intend to open ourselves up, release a documentation of the interface and co-develop new solutions together with you. In addition, we are planning to provide you with useful information around our heating systems you need for your development and to offer you a first-level support concerning our APIs.
>   
> As one of the users who is already working with our API, we would like to invite you to an early access to the Developer Portal. If you are interested, please fill out this short form. The access to the pre-version of the Portal will be sent to you within the next weeks.
>   
> We hope you understand the decision that we have to take. If you are interested to support us, on further development on the API and to explore new possibilities, we are very happy to welcome you to our Developer Portal soon!
>   
> In the meantime in case you have any questions, feel free to contact us under developer@viessmann.com.
>   
> Thank you for your understanding.
>   
> All the best!
>   
> Your Viessmann Developer Portal Team

I wouldn't hold out hope for the 'developer portal' as it's probably just a way to monetize the API.

ddokupil:
It'd be even ok as long as it's well documented and fair priced. The question is if it's going to happen any soon or they will disable access without enabling developer api.
Remember we can still use https://github.com/openv, but obviously it is way less convenient.

Could we also opt for access to pre-version?

oischinger:
https://www.viessmann-community.com/t5/Experten-fragen/Q-amp-A-Viessmann-API/qaq-p/127660

somm15:
I did upgrade the pip package .... but rate limiting is now in place.
So it doesn't work anymore. 
F**** Viessmann

Issue #22, opened 2020-02-09
Batching requests to the Viessmann servers


oischinger:
I was wondering if it would be possible to avoid doing too many requests at the same time. Currently we have a situation that the Home Assistant integration gets a lot of data via the API which result in lots of GET requests on the API at the same time.
With today's outage of the viessmann servers this results in lots of exceptions but it's also not nice to the servers.

Could we change/extend the API to select the data we're interested in and then requesting it with a single GET?

Would be interesting to get our thoughts on this.

crazyfx1:
Seems like they limited the amount of API requests now to 1450 per day.

```JSON
{
    "statusCode": 429,
    "errorType": "RATE_LIMIT_EXCEEDED",
    "message": "API calls rate limit has been exceeded. Please wait until your limit will renew.",
    "extendedPayload": {
        "name": "ViCare day limit",
        "requestCountLimit": 1450,
     }
}
```

crazyfx1:
I would suggest a method which executes the request and caches the JSON result. All other get methods use the cached JSON. 
In home-assistant the implementation is quite easy then: On each update, execute the request first and use the methods like before. And make sure we use a scan_interval of at least 60 seconds.

adorobis:
So we now have some data on the limits published on the forum [LINK](https://www.viessmann-community.com/t5/Experten-fragen/Betreff-Q-amp-A-Viessmann-API/qaq-p/127732/comment-id/117660#M117660):
```
Currently, we have two limits active:
120 calls for a time window of 10 minutes
1450 calls for a time window of 24 hours
We see these limitations reasonable, also based on your great explanation concerning cloud based services. So thank you for that!
Also, we decided against HATEOAS as it is deprecated and will sooner or later be switched off.
```
Any chance to update the integration to respect those limits?

oischinger:
I created a PR to Home assistant which sets the scan_interval of the ViCare integration to 15 minutes. It seems like we're on the safe side with that.
It should make it into HA 0.107.

But we could reduce the scan_interval quite a bit if we work on this.

crazyfx1:
The fix works for now, thanks.

Now we can discuss on how to implement the batching. Are there any plans so far @somm15 ?

somm15:
I think this has been solved.

PR #28, opened 2020-03-20
Proposal for batching requests

crazyfx1:
To overcome the Viessmann API limits I implemented a proposal for batching (issue #22)

I noticed that the normal call to the "features" endpoint already delivers all data, so there is no need to call each feature on its own.

Example:
```Python
t=HeatPump("xxx","xxx","token.save")
t.fetchAndCacheFeatures()
print(t.getDomesticHotWaterConfiguredTemperature()) 
```

Returns the same result as before.
Naming can be improved, but I couldnt come up with something else.

@oischinger Also looking forward to your feedback ;)

### DIFF
--- a/PyViCare/PyViCareCachedService.py
+++ b/PyViCare/PyViCareCachedService.py
@@ -1,0 +1,24 @@
+from datetime import datetime
+
+from PyViCare.PyViCareService import apiURLBase, ViCareService
+
+class ViCareCachedService(ViCareService):
+    
+    def __init__(self, username, password, cacheDuration, token_file=None,circuit=0):
+        ViCareService.__init__(self, username, password, token_file, circuit)
+        self.cacheDuration = cacheDuration
+        self.cache = None
+        self.cacheTime = None
+
+    def __setCache(self):
+        url = apiURLBase + '/operational-data/installations/' + str(self.id) + '/gateways/' + str(self.serial) + '/devices/0/features/'
+        self.cache = self.get(url)
+        self.cacheTime = datetime.now()
+
+    def getProperty(self,property_name):
+        if self.cache is None or self.cacheTime is None or (datetime.now() - self.cacheTime).seconds > self.cacheDuration:
+            self.__setCache()
+        
+        entities = self.cache["entities"]
+        feature = next((f for f in entities if f["class"][0] == property_name and f["class"][1] == "feature"), {})
+        return feature
--- a/PyViCare/PyViCareDevice.py
+++ b/PyViCare/PyViCareDevice.py
@@ -119,6 +119,9 @@ class Device:
     def deactivateComfort(self):
         return self.deactivateProgram("comfort")
 
+    def fetchAndCacheFeatures(self):
+        self.service.fetchAndCacheFeatures()
+
     def getMonthSinceLastService(self):
         try:
             return self.service.getProperty("heating.service.timeBased")["properties"]["activeMonthSinceLastService"]["value"]
@@ -172,7 +175,6 @@ class Device:
             return self.service.getProperty("heating.circuits." + str(self.service.circuit) + ".heating.curve")["properties"]["slope"]["value"]
         except KeyError:
             return "error"
-
 
     def getActiveProgram(self):
         try:
--- a/PyViCare/PyViCareDevice.py
+++ b/PyViCare/PyViCareDevice.py
@@ -24,7 +24,7 @@ class Device:
     """
 
     # TODO cirtcuit management should move at method level
-    def __init__(self, username, password,token_file=None,circuit=0):
+    def __init__(self, username, password,token_file=None,circuit=0,service=None):
         """Init function. Create the necessary oAuth2 sessions
         Parameters
         ----------
@@ -37,7 +37,10 @@ class Device:
         -------
         """
 
-        self.service = ViCareService(username, password, token_file, circuit)
+        if service is None:
+            self.service = ViCareService(username, password, token_file, circuit)
+        else:
+            self.service = service
 
     """ Set the active mode
     Parameters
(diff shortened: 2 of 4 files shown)


oischinger:
Nice finding @crazyfx1 . I'm testing it right now in combination with the Home Assistant integration and it seems to work as expected. 
From a library perspective I would however expect that people might be confused calling getProperty and never getting any new values.
Maybe we should call it getPropertyCached then and keep the existing getProperty as is?

crazyfx1:
Good point. But in my opinion the library was never playing fair with the API servers so I personally would not provide the old way of fetching data anymore. Maybe the owner should decide ;)

adorobis:
great @crazyfx1 !
Have you also made updates to the HA integration to use the new function?

oischinger:
The adaptions are straight forward. Just call t.fetchAndCacheFeatures() as the first thing in the climate and water heater component.

I'm currently testing it with a 40 seconds update interval. Let's see if I get banned before 24h ;)

paradix:
> The adaptions are straight forward. Just call t.fetchAndCacheFeatures() as the first thing in the climate and water heater component.
> 

I have been running this since 10 am this morning with a 60 sec interval. So far so good ;)

@oischinger does adding t.fetchAndCacheFeatures() to water and climate causes the api call to be done twice each 60 sec?

oischinger:
@paradix:

Yes it does. That's not optimal on the other hand it might still suffice.
What should be considered is that some people have multiple climate devices which would have to be perform different api calls anyway.
Therefore I believe a configurable scan interval is necessary for those not-so-common installations

paradix:
@oischinger is it possible to make the scan interval somehow configurable?

crazyfx1:
True I haven't thought about the water heater doing it's own API calls.
Since you only have 1450 calls per 24h (which is 1 per minute) you can't have both water heater and climate with scan_interval of 60.

Maybe we need another solution which shares the cache between PyViCareService instances.

paradix:
After about 14h of working fine with the 60s interval for climate and a 900s interval for water I got blocked by the api

oischinger:
Same here. I have my doubts about the 1450 calls limit...
But maybe I already used up lots of calls from the 24h period before switching to this pr

crazyfx1:
Other calls might are also be counted:
- The "get-installations" call (i'm not sure how often thats called actually)
- Setting temperatures 
- Getting new access tokens?

adorobis:
> After about 14h of working fine with the 60s interval for climate and a 900s interval for water I got blocked by the api

There must have been some disruption to the viessmann service, mine was not working in the morning neither and now is working again. I still have only the version with polling every 15 minutes

adorobis:
> Same here. I have my doubts about the 1450 calls limit...
> But maybe I already used up lots of calls from the 24h period before switching to this pr

But isn't the HA adding always 2 devices (the climate and water heater) by default? At least this is happening in my case. It will have 2 API calls every minute or am I misunderstanding something?

crazyfx1:
> > After about 14h of working fine with the 60s interval for climate and a 900s interval for water I got blocked by the api
> 
> There must have been some disruption to the viessmann service, mine was not working in the morning neither and now is working again. I still have only the version with polling every 15 minutes

I dont think the 15min version can even run 24h. The climate device alone does 16 requests per update and the water heater 3, so we have too many requests in the current version aswell ;)

paradix:
> > After about 14h of working fine with the 60s interval for climate and a 900s interval for water I got blocked by the api
> 
> There must have been some disruption to the viessmann service, mine was not working in the morning neither and now is working again. I still have only the version with polling every 15 minutes

Yes that't true, mine is back working again without the 24h penalty.



> > > After about 14h of working fine with the 60s interval for climate and a 900s interval for water I got blocked by the api
> > 
> > 
> > There must have been some disruption to the viessmann service, mine was not working in the morning neither and now is working again. I still have only the version with polling every 15 minutes
> 
> I dont think the 15min version can even run 24h. The climate device alone does 16 requests per update and the water heater 3, so we have too many requests in the current version aswell ;)

With the fetchAndCacheFeatures() method it should run without any trouble

crazyfx1:
@oischinger I've changed the API now to have the service cache data itself. So both water_heater and climate get to use the same cache.

Usage:
```Python
scan_interval = 60
t= HeatPump("xxx","xxx","token.save", 0, scan_interval)
print(t.getDomesticHotWaterConfiguredTemperature()) 
```

The only downside I see from this approach:
* It can occur that the cache is refreshed within one update-call in home-assistant. So you might see data which is not from the same api-result.

adorobis:
> The adaptions are straight forward. Just call t.fetchAndCacheFeatures() as the first thing in the climate and water heater component.
> 
> I'm currently testing it with a 40 seconds update interval. Let's see if I get banned before 24h ;)

I've just implemented those changes on my HA but I've set it up with scan_interval 120. So far so good, let's see in 24 hours. Many thanks for quick solution!

oischinger:
@somm15 What do you think of this implementation.
The last commit is imho a quite elegant solution. I would introduce a scan_interval parameter in the Home Assistant integration. Let's see if it survives 24hours with a scan_interval of 60 seconds, then we could set that as a default.

somm15:
Hi Guys,

I'll have a look tomorrow. Lock down with 4 kids isn't easy, so I'm a little bit slow for the moment.

somm15:
So after a short review, I will merge but I'll need to refactor later. 
The device ID is still hardcoded to 0 in the PyViCareSerice (and its new child). 
So far nobody never requested multi device support but the current design is not good.

I think we should have:
- Installation
  - Gateway(s)
    - Device(s)
Then I will probably move the GetFeature to the device object.

oischinger:
Thanks @somm15 for the merge and pypi release. I'm also slowed down by the corona lockdown of my family but thanks to @crazyfx1 I think we already have a reliable solution now.

I'll update the vicare custom component today evening and I'll contribute an update to Home Assistant Core as soon as we have a good idea what is a good value for the update frequency.

PR #29, opened 2020-03-21
Add oil burner support

o00batman00o:
I have created OilBoiler class based on GazBoiler class. Non-existing properties have been removed.
My goal is to integrate it in the home assistant integration to monitor burner hours.


crazyfx1:
Why are you not just using the GazBoiler, which is already in home-assistant? It already contains all existing properties of the oil boiler?

o00batman00o:
Yes you're right I didn't realize that home assistant does not get gaz specific properties.
As we can now get all properties in one request, it will be fast and easy to get more properties in home assistant.
To avoid getting errors on oil boiler with gaz specific properties,I think it is clearer to have a dedicated class.

PR #31, opened 2020-03-28
Add current desired dhw temperature

### LINKED ISSUE #24: water temp error
If I set the temperature from the app Vicare with different settings like in the photo attached,
I receive always the "normal" temp instead of the "temp2" value

photo_2020-02-12_12-24-22

I'm showing this issue in home assistant

min_temp: 10
max_temp: 60
current_temperature: 46
**temperature: 45**
target_temp_high: null
target_temp_low: null
friendly_name: ViCare Water
supported_features: 1

crazyfx1:
Add current desired dhw temperature by parsing the schedule.

Fixes #24 

Unfortunately the API does not provide any information on which temperature mode is currently active. So we have to parse the schedule to get the current mode.

When there are two scheduled temperature modes active at the same time, the "temperature 2" overrides all other modes according to the documentation of vitotronic.

@oischinger: Would you mind trying this out aswell? My dhw schedule is pretty basic so any other tests would be great ;) Also if this is approved, the HA integration should use the new property as target temperature in water_heater instead of the current configured temperature.

### DIFF
--- a/PyViCare/PyViCareDevice.py
+++ b/PyViCare/PyViCareDevice.py
@@ -302,6 +302,19 @@ class Device:
     def setDomesticHotWaterTemperature(self,temperature):
         return self.service.setProperty("heating.dhw.temperature","setTargetTemperature","{\"temperature\":"+str(temperature)+"}")
         
+    """ Set the target temperature 2 for domestic host water
+    Parameters
+    ----------
+    temperature : int
+        Target temperature
+
+    Returns
+    -------
+    result: json
+        json representation of the answer
+    """
+    def setDomesticHotWaterTemperature2(self,temperature):
+        return self.service.setProperty("heating.dhw.temperature.temp2","setTargetTemperature","{\"temperature\":"+str(temperature)+"}")
 
     def getCirculationPumpActive(self):
         try:
--- a/PyViCare/PyViCareDevice.py
+++ b/PyViCare/PyViCareDevice.py
@@ -2,11 +2,15 @@ import re
 import json
 import os
 import logging
+from datetime import datetime, time
 from PyViCare.PyViCareService import ViCareService
 from PyViCare.PyViCareCachedService import ViCareCachedService
 
 logger = logging.getLogger('ViCare')
 logger.addHandler(logging.NullHandler())
+
+VICARE_DAYS = ["mon", "tue", "wed", "thu", "fri", "sat", "sun"]
+VICARE_DHW_TEMP2 = "temp-2"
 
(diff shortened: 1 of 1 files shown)


crazyfx1:
@oischinger Also could you please tell me how to use an unpublished version of PyViCare in the custom component? :)

oischinger:
@crazyfx1 using an unpublished version depends a lot on your HA installation.
If you installed it on raspbian all you need to do is find the PyVicare code in the Home assistant python venv and modify it.

Are you using Home assistant OS (hass.io) then it's more complicated: You need to ssh to IP:22222 (which you cannot unless you provide your ssh key via an USB stick to the os), exec `login` and open a shell in the docker container:
```
docker exec -it homeassistant /bin/bash
```
You should be able to find the pyVicare files somewhere under  /usr/src/homeassistant/homeassistant

oischinger:
@crazyfx1 How can you actually set more than a single dhw temperature schedule? I thought you either use the heating schedule for dhw as well or you define schedule just for dh.

crazyfx1:
I'm not sure if I understand your question correctly. You can only have one dhw schedule, but you can have overlapping entries in that schedule.

For example a schedule might look like this
```JSON
"fri": [
    {
        "start": "18:00",
        "end": "22:00",
        "mode": "top",
        "position": 0
    },
    {
        "start": "18:00",
        "end": "19:00,
        "mode": "temp-2",
        "position": 1
    }
]
```

o00batman00o:
On my oil burner, I programmed an overlap on sunday with vicare application then read from pyViCare:
`{'active': True, 'mon': [{'start': '06:00', 'end': '22:00', 'mode': 'on', 'position': 0}], 'tue': [{'start': '06:00', 'end': '22:00', 'mode': 'on', 'position': 0}], 'wed': [{'start': '06:00', 'end': '22:00', 'mode': 'on', 'position': 0}], 'thu': [{'start': '06:00', 'end': '22:00', 'mode': 'on', 'position': 0}], 'fri': [{'start': '06:00', 'end': '22:00', 'mode': 'on', 'position': 0}], 'sat': [{'start': '06:00', 'end': '22:00', 'mode': 'on', 'position': 0}], 'sun': [{'start': '06:00', 'end': '22:00', 'mode': 'on', 'position': 0}, {'start': '08:00', 'end': '11:00', 'mode': 'on', 'position': 1}]}`

Maybe on newer boiler viessmann use 2 temperatures instead of on/off. Or they allow a "boost" temperature.

crazyfx1:
Yes seems like. But my implementation should work with both variants, since only "temp-2" is hardcoded.

PR #33, opened 2020-05-17
Add lock to PyViCareCachedService

crazyfx1:
We found an issue with the current implementation of the PyViCareCachedService while implementing additional sensors in HA.

Since the new sensors are running on different threads, it's possible that two threads enter the cache method at the same time. The get requests are then executed multiple times which result in bans from the viessmann API.

@somm15 It would be nice if you could create a new version of the library on PyPi after this merge. Thank you very much :)


adorobis:
Which are the new sensors? Is it already available in current release?

crazyfx1:
We created sensors out of existing attributes and added additional sensors. 
You can check the PR here: https://github.com/home-assistant/core/pull/34385
They are not available yet, otherwise every user would have problems with the API limit right now ;)

somm15:
@crazyfx1 Thanks for the contribution. New pip published. 
Regards.

PR #35, opened 2020-05-26
Fix issue for get Hotwater and Circulation Pump status

Nibot1:
I faced an issue while reading Pump staus of my Installation. In Vitoguide the Pumps are shown as active, but the code returns always that the Pumps where off. This code change fixed my issue.


Nibot1:
sorry found my issue

PR #45, opened 2020-11-01
add PyViCareFuelCell

nbraem:
Add new device class for the VitoValor PT2.
It inherits all properties from GazBoiler.

### DIFF
--- a/PyViCare/PyViCareFuelCell.py
+++ b/PyViCare/PyViCareFuelCell.py
@@ -1,0 +1,641 @@
+from PyViCare.PyViCareGazBoiler import GazBoiler
+
+class FuelCell(GazBoiler):
+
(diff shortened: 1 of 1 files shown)


somm15:
I don't really have a FuelCell so I can't check.
Are you sure your device has all the methods from the GazBoiler?
Should you inherit from Device instead?

nbraem:
Yes, the PT2 is a traditional gas boiler with a fuel cell. I tried all the methods from GazBoiler, they all work fine. If you want, I can share the complete json.

woehrl01:
@nbraem I just added some initial unit tests (see #55). It would be awesome if you could add and share your complete JSON, so we ensure that this works as expected in the future. If you don't have time you can send my the JSON, and I'll write some tests myself.

Issue #52, opened 2020-12-19
Change error in case of value not found


somm15:
@crazyfx1 and @oischinger 
As far as I know, you are the 2 main users and contributors. 

I change the error returned in PyVicareDevice from just "error" to 
except KeyError as err:
        	return "KeyError: "+str(err)

1) Is it a problem for the ha integration?

So it's already a little bit better. However, it's still far from clean. Error or not both return a string. 
I would prefer to raise a custom made error.
(or just stop stop catching the error and pass it, but I guess it would be more flexible to catch it and make an error custom made to have a message more clear). 

2) I guess small changes would be required in the HA integration to catch the error.

crazyfx1:
Yes this would indeed be an issue. The HA component checks if the return value is "error" and reacts accordingly.

I think it would be better to create and throw a custom exception (like PyViCareException) with the error text. Then the HA component needs to adapt and catch the exception.

oischinger:
I agree.
Actually custom exceptions is also what HA code style recommends for error handling.

PR #56, opened 2021-01-02
Feature: Raise exception instead of returning error string

woehrl01:
This PR changes the error handling as discussed in #52, to raise an `PyViCareNotSupportedFeatureError` if the feature is not supported.

The feature is hidden behind a feature flag, so that the old behavior can still be retained.

I also moved part of the raising of the error to the underlying logic, so that it's not triggered by another a `KeyError`. This should uncover some possible bugs. This should allow us also to remove the decorator all together (except for the `IndexError`). But we lack a few test samples to make this 💯 .

Lastly the PR fixes a circular dependency which fails the build, by removing the already deprecated `ViCareSession` (of course we could just move the decorator, but I think the cleanup is the better approach).

### DIFF
--- a/PyViCare/Feature.py
+++ b/PyViCare/Feature.py
@@ -1,2 +1,4 @@
-#feature flag to have a release which still returns an error string
+# Feature flag to raise an exception in case of a non existing device feature.
+# The flag should be fully removed in a later release. 
+# It allows dependend libraries to gracefully migrate to the new behaviour
 raise_exception_instead_of_returning_error = False
--- a/PyViCare/Feature.py
+++ b/PyViCare/Feature.py
@@ -1,4 +1,4 @@
 # Feature flag to raise an exception in case of a non existing device feature.
 # The flag should be fully removed in a later release. 
 # It allows dependend libraries to gracefully migrate to the new behaviour
-raise_exception_instead_of_returning_error = False
+raise_exception_on_not_supported_device_feature = False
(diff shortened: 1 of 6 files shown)


somm15:
You're doing all the job, amazing! Thanks.
Sorry but I'm done for today, I'll review tomorrow.

woehrl01:
No worries, Simon! Take your time! 😄 Thanks for your great work!

oischinger:
Amazing job guys!
I guess we should move to exceptions instead of error strings wherever possible. Of me course we need to adapt the Home Assistant integration.

Maybe one request: Would it be possible to release the PyVicare library again once before we migrate to exception handling?
This would allow us to fix this issue here first: https://github.com/somm15/PyViCare/commit/e1cce406c05f72ce43a2aa9c0c97552755f25b09

I believe so far all changes should still be compatible, correct?

woehrl01:
Sure @oischinger, but let me please split this PR up, so that we have a version where all the tests pass. 

edit: changed my mind 😄 , please read further why.

woehrl01:
@oischinger @somm15 I decided against a second PR and changed this feature to be hidden behind a feature flag (`PyViCare.Feature.raise_exception_on_not_supported_device_feature`). This ensures a graceful adoption of dependent libraries. 
The flag is disabled by default for the next release and should be enabled by default in a subsequent release.

woehrl01:
@oischinger to your last question. Yes, all changes are compatible, besides the removing of the deprecated `ViCareSession` type, but to my knowledge you don't use that in the HA integration. Despite that, it just fixes some bugs.

PR #62, opened 2021-01-04
Add feature to support one time charge of dhw

woehrl01:
This PR adds support for "One Time Charge". 

Some devices have the feature "One Time Charge" to heat the water a single time out of schedule.

### DIFF
--- a/PyViCare/PyViCareService.py
+++ b/PyViCare/PyViCareService.py
@@ -26,6 +26,12 @@ def readFeature(entities, property_name):
     feature = next((f for f in entities if f["class"][0] == property_name and f["class"][1] == "feature"), {})
     return feature
 
+def buildSetPropertyUrl(id, serial, property_name, action):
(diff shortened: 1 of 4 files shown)


PyViCare/PyViCareService.py
```diff
@@ -238,12 +244,12 @@ def get(self, url):
 
    #TODO should move to device after refactoring 
     def getProperty(self,property_name):
-        url = apiURLBase + '/operational-data/installations/' + str(self.id) + '/gateways/' + str(self.serial) + '/devices/0/features/' + property_name
+        url = buildGetPropertyUrl(self.id, self.serial, property_name)
```
somm15, line 252: Why did you get rid of the 404 error handling?

somm15 (approved):

PyViCare/PyViCareService.py
somm15: Good side is that you added that anyway. I'll merge and test.

woehrl01: What do you mean with 404 handling? I just extracted the URL concatenation, so I can use the same logic inside the service mock.

PR #75, opened 2021-02-28
add FuelCell tests

nbraem:


nbraem:
Note that it only tests the fields in FuelCell, although FuelCell inherits from GazBoiler. This is because it fails for the GazBolier fields listed below. When I added FuelCell a few months ago, these still worked. So someone should check with a regular GazBoiler if these are still supported.
```
getBoilerTemperature
getDesiredTemperatureForProgram
getDomesticHotWaterChargingActive
getDomesticHotWaterConfiguredTemperature2
getDomesticHotWaterPumpActive
getLastServiceDate
getMonthSinceLastService
getPrograms
getRoomTemperature
```

nbraem:
It also looks like tests are not running on PRs 😞

PR #82, opened 2021-05-28
Fix heatpump properties

crazyfx1:
Viessmann changed properties for heatpumps.
Previous the data was available in `heating.compressor` but now it's indexed in `heating.compressors.0` (for first circuit).


nobwyn:
Any update on this? 

I'm using Home Assistant's  ViCare integration, which ist based on this library and also have now the same issue at the moment... 

Would be great to see the library update in near future 😏

crazyfx1:
We need @somm15 to approve the PR and create a new version ;)

TvoDebug:
The fix from @crazyfx1 works great. Would be also nice to have a hybrid class which is combining gas and heatpump. At the moment I'm realizing it by two different requests, which leads to halving the limit per day.

crazyfx1:
Are you using home-assistant or PyViCare standalone?
If standalone you can create a PyViCareCachedService and pass it to both instances of gas and heatpump.

idstein:
Fix works. I needed to create a custom_component overwrite in homeassistant to get the readings again.

PR #88, opened 2021-06-19
Add support for V2 API

crazyfx1:
This PR adds support for the Viessmann V2 API. Instructions how to use the new API are in the readme.

The most important part is that every user has to create an account in the viessmann developer portal and use their personal client id.

### DIFF
--- a/PyViCare/PyViCareServiceV2.py
+++ b/PyViCare/PyViCareServiceV2.py
@@ -34,10 +34,10 @@ def readFeature(entities, property_name):
     return feature
 
 def buildSetPropertyUrl(id, serial, circuit, property_name, action):
-    return apiURLBase +'/operational-data/v1/installations/'+str(id)+'/gateways/'+str(serial)+'/devices/'+str(circuit)+'/features/'+property_name+'/'+action
+    return apiURLBase +'/equipment/installations/'+str(id)+'/gateways/'+str(serial)+'/devices/'+str(circuit)+'/features/'+property_name+'/'+action
 
 def buildGetPropertyUrl(id, serial, circuit, property_name):
-    return apiURLBase + '/operational-data/installations/'+str(id)+'/gateways/'+str(serial)+'/devices/'+str(circuit)+'/features/'+property_name   
+    return apiURLBase + '/equipment/installations/'+str(id)+'/gateways/'+str(serial)+'/devices/'+str(circuit)+'/features/'+property_name   
 
 # https://api.viessmann-platform.io/general-management/v1/installations/DDDDD gives the type like VitoconnectOptolink
 # entities / "deviceType": "heating"
--- a/PyViCare/PyViCareServiceV2.py
+++ b/PyViCare/PyViCareServiceV2.py
@@ -1,0 +1,292 @@
+from requests_oauthlib import OAuth2Session
+from oauthlib.oauth2 import TokenExpiredError 
+
+import requests
+import re
+import pickle
+import os
+import logging
+import datetime
+import pkce
+from pickle import UnpicklingError
+# This is required because "requests" uses simplejson if installed on the system 
+
+import simplejson as json
+from simplejson import JSONDecodeError
+from PyViCare.PyViCare import PyViCareNotSupportedFeatureError, PyViCareRateLimitError
+import PyViCare.Feature
+
+authorizeURL = 'https://iam.viessmann.com/idp/v2/authorize'
+token_url = 'https://iam.viessmann.com/idp/v2/token'
+apiURLBase = 'https://api.viessmann.com/iot/v1'
+redirect_uri = "vicare://oauth-callback/everest"
+viessmann_scope=["IoT User"]
+logger = logging.getLogger('ViCare')
+logger.addHandler(logging.NullHandler())
+
+
+def readFeature(entities, property_name):
+    feature = next((f for f in entities if f["feature"] == property_name), None)
+
+    if(feature is None):
+        raise PyViCareNotSupportedFeatureError(property_name)
+
+    return feature
+
+def buildSetPropertyUrl(id, serial, circuit, property_name, action):
+    return apiURLBase +'/operational-data/v1/installations/'+str(id)+'/gateways/'+str(serial)+'/devices/'+str(circuit)+'/features/'+property_name+'/'+action
+
+def buildGetPropertyUrl(id, serial, circuit, property_name):
+    return apiURLBase + '/operational-data/installations/'+str(id)+'/gateways/'+str(serial)+'/devices/'+str(circuit)+'/features/'+property_name   
+
+# https://api.viessmann-platform.io/general-management/v1/installations/DDDDD gives the type like VitoconnectOptolink
+# entities / "deviceType": "heating"
+# entities are connected devices
+# https://api.viessmann-platform.io/general-management/v1/installations/16011/gateways PUT and POST only
+
+# TODO handle multi install / multi devices
+
+""""Viessmann ViCare API Python tools"""
+
+class ViCareServiceV2:
+    """This class connects to the Viesmann ViCare API.
+    The authentication is done through OAuth2.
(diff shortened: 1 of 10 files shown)


PyViCare/PyViCareServiceV2.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # TODO tranform to exception
+        
+    def renewToken(self):
+        logger.info("Token expired, renewing")
+        self.oauth=self.__getNewToken(self.username,self.password,self.token_file)
+        logger.info("Token renewed successfully")
+            
+        
+    """Get URL using OAuth session. Automatically renew the token if needed
```
hnykda, line 185: A quite random note, feel free to ignore, but it's pretty weird for python to see a docstring above a function name.

adorobis:
Hi, many thanks for your work on the v2 API! I'm trying to test it but I'm receiving ERROR output for any API call when using this version. The authentication must work ok because I can see that the call was made in the statistics on the viessmann dev portal. When testing the API in postman all works ok. Now after looking in more detail all the get requests from the test pack are ending up with error:
```
{
    "viErrorId": "req-47a5908e7cac4488b1089841638632b8",
    "statusCode": 404,
    "errorType": "ENDPOINT_NOT_FOUND",
    "message": "Endpoint /iot/v1/operational-data/installations/xxxx/gateways/yyyyy/devices/0/features/heating.sensors.temperature.outside does not match any existing route"
}
```
My device is vitodens 200-W.

And by looking at all the features that my device exposes over the API the outside temperature url should be:
`https://api.viessmann-platform.io/iot/v1/equipment/installations/xxxx/gateways/yyyyyy/devices/0/features/heating.sensors.temperature.outside`
By briefly looking into it it seems that the `operational-data` should be `equipment` in case of my device but the returned data is in a different format than expected, for this parameter I'm getting:
```
{'data': {'properties': {'unit': {'value': 'celsius', 'type': 'string'}, 'value': {'type': 'number', 'value': 28.5, 'unit': 'celsius'}, 'status': {'type': 'string', 'value': 'connected'}}, 'commands': {}, 'components': [], 'apiVersion': 1, 'uri': 'https://api.viessmann-platform.io/iot/v1/equipment/installations/xxxx/gateways/yyyyyy/devices/0/features/heating.sensors.temperature.outside', 'gatewayId': 'yyyyyy', 'feature': 'heating.sensors.temperature.outside', 'timestamp': '2021-06-22T09:16:03.204Z', 'isEnabled': True, 'isReady': True, 'deviceId': '0'}}
```
whereas the `getOutsideTemperature` function is expecting `getProperty("heating.sensors.temperature.outside")["properties"]["value"]["value"]`

crazyfx1:
Hi @adorobis ,

thanks for your hint. The service indeed had the wrong URL. I didn't realize because I'm always using the cached version ;)

Ok seems like the call for single features is different now and returns a single property instead of a list. Please try the cached version in the meantime.

adorobis:
@crazyfx1 
Many thanks for quick reaction. Indeed the cached service works much better. Out of the examples provided in the testing suggestions here are my results:
```
print(t.getDomesticHotWaterConfiguredTemperature())
44
print(t.getDomesticHotWaterStorageTemperature())
44.3
print(t.getOutsideTemperature())
28.5
print(t.getRoomTemperature())
24.6
print(t.getSupplyTemperature())
52
print(t.getHeatingCurveShift())
error
print(t.getHeatingCurveSlope())
error
print(t.getBoilerTemperature())
52
print(t.getActiveProgram())
standby
print(t.getPrograms())
error
print(t.getCurrentDesiredTemperature())
error
print(t.getMonthSinceLastService())
error
print(t.getLastServiceDate())
error
print(t.getActiveMode())
dhw
print(t.getDesiredTemperatureForProgram("comfort"))
22
print(t.setProgramTemperature("comfort",21))
{'viErrorId': 'req-8f90f4a420224fd5b46f37a5abcdc7ad', 'statusCode': 404, 'errorType': 'RESOURCE_NOT_FOUND', 'message': 'Required resource was not found', 'extendedPayload': {}}
print(t.activateProgram("comfort"))
{'viErrorId': 'req-dd3cf45973dc49e9aa440ece2d637a86', 'statusCode': 404, 'errorType': 'RESOURCE_NOT_FOUND', 'message': 'Required resource was not found', 'extendedPayload': {}}
print(t.setDomesticHotWaterTemperature(42))
{'viErrorId': 'req-49e08cb153ee4aa5aa9d8cfe314afb17', 'statusCode': 404, 'errorType': 'RESOURCE_NOT_FOUND', 'message': 'Required resource was not found', 'extendedPayload': {}}
print(t.deactivateComfort())
{'viErrorId': 'req-d023a6d9a89348ee92eb59e28ce2077a', 'statusCode': 404, 'errorType': 'RESOURCE_NOT_FOUND', 'message': 'Required resource was not found', 'extendedPayload': {}}
```
So obviously some services don't work the same way in v2. Let me know if you'd like me to get into more details for each of the functions which end up with error.
Most likely we can ignore the `setProgramTemperature("comfort",21)`, `activateProgram("comfort")` and `deactivateComfort()` as those didn't work for my heater in v1 API either.

crazyfx1:
It would be great if you could provide a PR which contains your whole V2 result as json-file in the tests-folder. We can add the tests later, but this helps to see which properties are not working anymore.
Make sure to remove your installationId, serial and deviceId from the result.

adorobis:
Here it is. Let me know if this is what you've expected,
https://github.com/somm15/PyViCare/pull/89

crazyfx1:
Yes thank you very much. As soon as I'm maintainer I can approve it ;)

jeti-flight:
Hi, just found this AFTER my merge request #90 using Authlib insted of oauth.
Attached a diff style file with the missing data points. First column is v1 API, second the new API. 74 data points are missing. Just wrote to viessmann about this.
[feature-diff.txt](https://github.com/somm15/PyViCare/files/6704965/feature-diff.txt)

crazyfx1:
Thanks. 

My answer from viessmann was, that they don't have plans to add the missing datapoints, maybe in the future.

adorobis:
> Hi, just found this AFTER my merge request #90 using Authlib insted of oauth.
> Attached a diff style file with the missing data points. First column is v1 API, second the new API. 74 data points are missing. Just wrote to viessmann about this.
> [feature-diff.txt](https://github.com/somm15/PyViCare/files/6704965/feature-diff.txt)

Hi, looking at your PR I'm afraid that none (or just many?) of the features will work as the new API has a different base URLs. Look at this commit:
https://github.com/crazyfx1/PyViCare/commit/238f4df2b8aed565f0e6d974752e9722c498b569
Btw, how are you testing the API?

jeti-flight:
Hi,
in my "new" PyViCare files the url's are adjusted to the new url's. I've run my version against my own Vitodens300 over 24h collecting all existing data with my own "single-request.py" not the individual data point access (this will go beyond the 1440 limit per day if you are doing this every 5mins, seems getting ALL data every 5mins in one shot counts only once!). Doing this btw. with the now "old" API since May. I'm sure the new version works as i can see including collecting installationId, gatewaySerial, build the first token and renew it. I have done no regression test for the individual data points or seen any limit errors. Have done some checks with postman as well looking at the differences old<>new re data points. The diff attached shows only missing things no changes. The complicated part was to understand the oauth with pkce and Viessmann's documentation. Postman was helping a lot (see the new readme), the rest was try and error. With the missing data points for the near future or forever it becomes a little less useless, shame on Viessmann. Must be a way to collect as the App potentially will show you the consumptions etc. Let's wait until then.

jackie-t:
Just received another mail from Viessmann, stating that the old api will be disabled tomorrow, which probably means that PyViCare will cease to function.

Is there an easy way to get the new version of PyViCare? pip install --upgrade  states I'm on the newest version. Since the old version will be completely useless in a bit, would it make sense to push whatever you have to test out so the broad user base can get it?

PR #90, opened 2021-06-23
V2 devel

jeti-flight:
Hi, was using this toll since my new heating system bwecomes online back in may. Now after getting the nice mail from Viessmann about the API change i was trying to slove the dilemma. See this please as a first step.
Changes:
- implement a single request for all feature to save the limit and pull all interesting values by hand
- discover that the consumption values are restet accourding UTC (not sure why but during CEST 2h to late...)
- changing the limit handling a little
- start using Authlib after this was done
- implementing Oauth with PKCE
- adjusting the single request
- some values aren't anymore int the list (consumption, pressure, modulation, requested feedback from Viessmann just today, no answer)

I hope i understood you logic the right way.
(I'm new to Github, so please be patient....)


crazyfx1:
Thanks for your contribution. Unfortunately this PR is very big and there are a few issues with it:
* The V2 API was already implemented. If you believe your implementation has advantages over mine, please feel free to create a new PR containing the changes.
* The current version already can already cache the results to limit API calls. The functionality seems to be very hidden though, since you are not the first one to miss it.
* Calling sys.exit is very bad, since this library is also used in Home-Assistant. Your call would terminate the process.

Can you please review your changes and provide smaller portions so they can be integrated easier.

PR #98, opened 2021-07-13
Add api for solar collector

sweh:
Add api methods to get the current solar panel temperature and if the solar pump is running.

### DIFF
--- a/PyViCare/PyViCareDevice.py
+++ b/PyViCare/PyViCareDevice.py
@@ -358,5 +358,5 @@ class Device:
 
     @handleNotSupported
     def getSolarPumpActive(self):
-        status =  self.service.getProperty("heating.solar.pumps.circuit")["properties"]["status"]["value"]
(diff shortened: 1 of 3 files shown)


crazyfx1:
Thanks for your contribution.

Are those calls tested with the V2 API?

sweh:
Tested under V1 and V2.

crazyfx1:
It would be great if you could add your V2 json response to the tests folder and add unit tests for the new properties.
We only need the V2 version, the V1 will stop working in 2 days. Also please don't forget to remove your gatewaySerial and installationId from the response.

sweh:
> It would be great if you could add your V2 json response to the tests folder and add unit tests for the new properties.
> We only need the V2 version, the V1 will stop working in 2 days. Also please don't forget to remove your gatewaySerial and installationId from the response.

I extended the existing reponses json for a Vitodens 200W (which is the one I own) and added two lightweight test methods to the existing ones.

PR #100, opened 2021-07-14
features response update after new parameters added by viessmann

adorobis:
Here is the latest response to features request on the V2 API


woehrl01:
Thank @adorobis . I'd suggest replacing the serial, gateway id and installationid with some random text.

adorobis:
Thanks @woehrl01 , updated now. I've tried but can't remove the history of it from the pull request...

woehrl01:
No worries I do a sqash merge so the history is removed in the main line. Thank you for the response file.

PR #102, opened 2021-07-14
Flip feature flags to raise exceptions on certain errors

woehrl01:
As we flip the major version, this PR flips the feature flags. We should still keep them there so that dependent libraries can flip it back.


crazyfx1 (approved):
Yes, good idea :)

PR #112, opened 2021-07-19
Create response_Vitodens222W.json

paradix:


jborup:
@paradix would it be possible for you to extract this dataset again?
Looks like it is extracted with ' rather than ".
And also all booleans are without double quoutes("). Of course this can be fixed, but would suggest to pull a new extract if possible.

paradix:
I hope now it formatted in the right way.

jborup:
@paradix looks perfect now thanks :-)

jborup:
@woehrl01 would you be able to do the merge on this one and then I will rename and add the test.

Issue #115, opened 2021-07-22
Investigate HA OAuth2 facilities


oischinger:
Home Assistant has built-in support for integrations that offer account linking using the OAuth2 authorization framework. To be able to leverage this, you will need to structure your Python API library in a way that allows Home Assistant to be responsible for refreshing tokens.

https://developers.home-assistant.io/docs/config_entries_config_flow_handler/#configuration-via-oauth2

To be honest I have no idea about all this OAuth stuff. Does someone here have enough knowledge to judge if this is feasible?
Would certainly be a nice feature.

woehrl01:
This shouldn't be such a big issue, we just need to extract/refactor the authentication. See: https://developers.home-assistant.io/docs/api_lib_auth

I can get my hands dirty over the weekend.

PR #119, opened 2021-07-25
Refactor OAuth to pass an external implementation

woehrl01:
PR to be able to replace the OAuth dealing with external logics, e.g. Home Assistant (#115)

Changes:

* Added abstract implementation which wraps the local OAuth handling
* Added a new facade class `PyViCare` to interact with your devices. With this class, you can iterate over all your existing devices. See the changed README for usage description
* Added `ViCareHomeAssistantOAuthManager` as an example, which of course should be moved to the Home Assistant part.
* Added name of the device and online status output.
* Added an integration test which automatically runs against your local device, you need to provide credentials for this. The tests can be run with `./run_test.sh` (you likely need to `chmod +x` first)
* Added a `./format.sh` script to have a consistent formatting over the code. Will create a PR after this one is merged.
* Added an `PyViCareInvalidCredentialsError` which gets thrown if there are invalid credentials provided and a token can't be fetched
* Some other minor refactorings and bugfixes, I came along throughout the major refactoring and can be grouped together.


crazyfx1 (commented):
The refactoring looks great!

I don't like the new interface though. I don't want the user to deal with the service instance.

In an ideal world I would imagine the usage somehow like this:
There is an PyViCare object with an init method to initialize the connection. Then everything is already filled.
```Python
pyvicare = PyViCare()
pyvicare.init(username, password, clientid)
# or 
pyvicare.init(oauthmanager) 
# pyvicare object is now populated
pyvicare.installations # contains installation
pyvicare.installations[0].gateways # contains gateways
pyvicare.installations[0].gateways[0].devices # contains devices
pyvicare.devices # contains all devices directly aswell
pyvicare.devices[0] # is already of type GazBoiler if you have a GazBoiler
```

Unfortunately we can't really tell which device we use, so the last step is not possible. But we could just implement
```Python
pyvicare.getGazBoilerDevice(0) # gets first device which user has to know is a gas boiler
```

For this to work the init method must already execute one data request to the API. But since oischinger wants to check which sensors are available on initialization we need that call anyways ;)

crazyfx1:
Ok you now changed to a builder pattern, which works ofc. I can imagine that a level higher:
```Python
pyvicare = PyViCareBuilder()
                     .withClientId(clientId)
                     .withCredentials(username, password)
                     .build()
```

woehrl01:
Thanks for your input. I guess we are on a good track for this, with the current implementation. I suggest creating a `PyViCare` class which simple wraps the builder and all the other types. 
Furthermore, I had the idea of adding the type to the builder, too. But that hasn't looked right. Your idea filled the gap, awesome.

edit: I just added an initial version of a `PyViCare` class. To make this work nicely, I need to do some heavier refactoring, so we can pass the serial, installation and circuit around more freely.
edit: done.

PyViCare/PyViCare.py
```diff
@@ -1,42 +1,67 @@
...
+
+    def __loadInstallations(self):
+        installations = self.oauth_manager.get("/equipment/installations?includeGateways=true")
+        self.devices = []
+        for installation in installations["data"]:
+            id = installation["id"]
+            for gateway in installation["gateways"]:
+                serial = gateway["serial"]
+
+                accessor = ViCareDeviceAccessor(id, serial, 0)
```
woehrl01: We still miss iterating through the devices.

PyViCare/PyViCareOAuthManager.py
```diff
@@ -0,0 +1,215 @@
...
+                apiURLBase + url, data, headers=headers).json()
+            self.handleExpiredToken(response)
+            self.handleRateLimit(response)
+            return response
+        except TokenExpiredError:
+            self.oauth_manager.renewToken()
+            return self.post(url, data)
+
+
+class ViCareHomeAssistantOAuthManager(AbstractViCareOAuthManager):
```
woehrl01, line 83: Move this to the HomeAssistant implementation

woehrl01:
@crazyfx1 Please have another look, I implemented the iteration over all devices. 

I thought about replicating the installations and gateways, but I don't know about a reasonable use case for that. As we are only accessing the devices, I created a list of all devices, like in your pseudocode. Each device has all necessary information stored in the `.config` property, for filtering, I'd suggest using that data.

crazyfx1:
I like the new interface very much. I wanted to try it out but I couldn't get it imported correctly. Maybe it's because the class is also called PyViCare just like the library and the module?

woehrl01:
Just renamed it to `ViCare` for now, not sure about if we get licensing issues here?

edit: found the issue. I'm about to fix it

crazyfx1:
Did you find the circular import?

woehrl01:
So, I fixed the cyclic dependency and added an integration test you can run with your own credentials. That means we are always on the safe spot for the new `PyViCare` class. The name is changed back to `PyViCare`, as it's not an issue.

crazyfx1:
image

The PyViCareCachedService is now broken :)

woehrl01:
😅 On my way... 

edit: done

crazyfx1:
Ok the access works. Only the properties with service.circuit in it are broken :)

woehrl01:
Thank you very much @crazyfx1 for testing this so in-depth on your side. I fixed the `circuit` reference. I also just stumbled over the issue if you have a misconfiguration of credentials. I added a meaningful exception in that case.

I also added a `format.sh` file, which we can run after everything, which should give us some cleaning. I didn't run it yet as it would bloat the diff pretty hard.

oischinger:
Oh boy, this is quite a huge refactoring.
Meanwhile I've started a config flow for the HA integration here: https://github.com/oischinger/ha_vicare/pull/43

This requires the classic credentials and client_id but we can modify it to use the HA oauth flow steps.

woehrl01:
@oischinger the refactoring looks bigger than it is. Most of the stuff is just moved around to other files. The testing part is a complete addon. The addition of the PyViCare class was needed to be able to replace the oauthmanager in a seemless way. It should make the usage from Home Assistant much easier, as we can now automatically detect all devices and also the device type based on the name of the device.

PR #126, opened 2021-08-02
Refactor circuits for auto detection and easier usage

woehrl01:
This PR is another breaking change. But it allows a much easier usage of all the circuits.

Changes:
1. Split up of the device in non-circuit and circuit dependent properties
2. Access the circuit properties via `device.circuits[0].getAbc()`

This allows iteration of all existing circuits via the new `.circuits` property. Circuits are accessed via an indexer.

### DIFF
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -54,6 +54,6 @@ class PyViCare:
                         installation_id, gateway_serial, device_id)
                     service = self.__buildService(accessor)
 
-                    logger.info("Device found: %s" % device_model)
+                    logger.info(f"Device found: {device_model}")
 
                     yield PyViCareDeviceConfig(service, device_model, status)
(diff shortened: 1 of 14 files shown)


oischinger:
looks good. Is there a way to retrieve the available circuits from the API?
This way Home assistant could just go through what's available instead of having the user to configure them.

woehrl01:
@oischinger Yes, that's already implemented `getActiveCircuits` I use that one to generate the backing list. That means you can simply iterate through all the `circuits`. I use that in the `test_Integration.py` 

To automatically iterate through all the available ones in Home Assistant was exactly my intention. What we need to modify in Home Assistant is that you can choose the circuit for `setMode`.

oischinger:
perfect. I guess we need another release after this PR is merged :D

woehrl01:
See `2.3`

PR #130, opened 2021-08-03
Adding Vitodens 300W data and test

jborup:
Please find a dataset from a Vitodens 300W and also the tests.

### DIFF
--- a/tests/response_Vitodens300W.json
+++ b/tests/response_Vitodens300W.json
@@ -1,0 +1,3080 @@
+{
+    "data": [
+        {
+            "apiVersion": 1,
+            "isEnabled": true,
+            "isReady": true,
+            "gatewayId": "yyyyyyyyyyyyyyyy",
+            "feature": "heating",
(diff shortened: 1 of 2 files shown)


woehrl01:
@jborup Thank you very much for the PR. I suppose that the code comments of the tests is because they are not supported for your device? Please change the tests so that they validate the not supported exception. After that, I'm happy to merge this in.

jborup:
@woehrl01 added the exception validations, but I left the "desired" outcome, as this is what Viessmann really should provide, so i hope we could end up with this not being unsupported later :-)

woehrl01:
Thanks @jborup as we already have tests in place for those attributes on the 200W where this is supported from Viessman, I, personally, prefer the comments to be removed. What you can do is to replace them with something like `Is currently (August, 2021) not supported by the Viessman API even though it works for the Vitodens 200W`. So, we remember that in the future.

jborup:
@woehrl01 Last update for vitodens300W for now - take it or leave it:-)

woehrl01 (approved):
Thank you! Looks great 🚀

PR #140, opened 2021-08-04
Adding features to extract test dataset and remove serial/installid information

jborup:
Suggestion to add a few functions and a recipe to ensure people can extract with prescriptive instructions in addition to doing postman and search and replace manually.

Happy to help out with adding the test cases along the line, when time permits :-)


woehrl01:
That looks really good! Thank you!

PR #141, opened 2021-08-04
Initial test for Vitodens333F

jborup:
Added test cases for Vitodens333F expanded with a few additional tests compared to other tests (for outside and boiler temperature).


woehrl01:
Awesome! Thank you @jborup

PR #146, opened 2021-08-05
Adding tests for Vitodens222W

jborup:
@woehrl01 as agreed tests now added for Vitodens222W as well.
I updated .gitignore to also ignore env* not just env, just makes it easier :-)


woehrl01:
Awesome! Thank you!

PR #148, opened 2021-08-06
Adding response Vitocal 300G heatpump

blehers:
Adding response Vitocal 300G


woehrl01:
Thank you @blehers. Thats great!

PR #175, opened 2021-08-24
Add dump for my Vitocaldens 222 F

kventil:
Please let me know if you need any additional information.

Thanks for your work! ❤️


woehrl01:
Thanks @kventil! 

I'll need to think about, how to best reflect a hybrid device. (If we should duplicate all the properties or if we should simply use multiple inheritance.) 

I need to compare the available attributes and craft a PR for the added new device afterwards.

PR #178, opened 2021-09-02
Breaking: Separate circuit, burner and compressor and add hybrid device

woehrl01:
We currently only "use" circuits. The problem is that there is a difference between `circuits`, `burners` and `compressors`. That means that we mix the property of all these up.

This PR fixes that, by introducing proper separation between these parts of the device.

* Adds hybrid device as discussed in #175 
* Add all missing properties, existing in response files
* Fixed `heating.power.consumption.total` due to API change
* Added handler for 500 status code errors (see #179)
* Added unittest for new properties


PyViCare/PyViCareGazBoiler.py
```diff
@@ -112,16 +122,20 @@ def getPowerConsumptionThisYear(self):
...
 
-class GazBoilerWithCircuit(DeviceWithCircuit):
+class GazBurner(DeviceWithComponent):
+
+    @property
+    def burner(self) -> str:
+        return self.component
 
     @handleNotSupported
     def getBurnerHours(self):
```
crazyfx1: We could think about renaming them to just "getHours()", "getStarts()".

PyViCare/PyViCareHeatPump.py
```diff
@@ -20,36 +30,40 @@ def getReturnTemperaturePrimaryCircuit(self):
...
 
-class HeatPumpWithCircuit(DeviceWithCircuit):
+class Compressor(DeviceWithComponent):
+
+    @property
+    def compressor(self) -> str:
+        return self.component
 
     @handleNotSupported
     def getCompressorStarts(self):
```
crazyfx1: Same here, "getStarts()", "getHours()"

crazyfx1 (approved):
Looks great :)

PyViCare/PyViCareHeatPump.py
woehrl01: great idea 👍🏼

PR #196, opened 2021-09-30
fix: nameing of getActive

woehrl01:
Follow up of PR #195, I chose the name `getIsActive` we already have the name `getActive` for compressor, so this is a change 
to keep this consistent.

### DIFF
--- a/PyViCare/PyViCareGazBoiler.py
+++ b/PyViCare/PyViCareGazBoiler.py
@@ -133,7 +133,7 @@ class GazBurner(DeviceWithComponent):
         return self.component
 
     @handleNotSupported
-    def getIsActive(self):
+    def getActive(self):
         return self.service.getProperty(f"heating.burners.{self.burner}")["properties"]["active"]["value"]
 
     @handleNotSupported
--- a/PyViCare/PyViCareOilBoiler.py
+++ b/PyViCare/PyViCareOilBoiler.py
@@ -5,7 +5,7 @@ from PyViCare.PyViCareUtils import handleNotSupported
 class OilBoiler(Device):
 
     @handleNotSupported
-    def getIsActive(self):
+    def getActive(self):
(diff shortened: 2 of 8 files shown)


SPITZGITHUB:
Hello,

I am a little confused about the different changes you have made. I tried all names I found in the discussion but none functioned: 
As shown in https://github.com/somm15/PyViCare
with 

**_vicare = PyViCare()
vicare.initWithCredentials(email, password, client_id, "token.save")
device = vicare.devices[0]
print(device)
t = device.asAutoDetectDevice()
burner = t.burners[0] #select burner
print(burner.getActive())_**

Result: 
**_burner = t.burners[0] #select burner
AttributeError: 'FuelCell' object has no attribute 'burners'_**

Can you give me the right naming behind the **_burner._** so I get the status of the burner for FuelCell?

woehrl01:
Thanks for reaching back out to me @SPITZGITHUB . I must have missed that your device was a fuel cell device. Can you tell me which device type you have? Furthermore may I use your result file to create tests for fuel cell, as we current lack an up to date response for that? That would be really helpful.

SPITZGITHUB:
Hallo 
in my code I sent it to you, some days before,  it was written:

device = vicare.devices[0]
**_t = device.asFuelCell()_**
circuit = t.circuits[0] #select heating circuit

so I thought it was clear that this is a fuel cell?
What do you mean with device type? It is a VITOVALOR PT2.
Sure you can take the results for your test cases. I can also support you in test if necessary. 

I also miss some possibilies for my device, e.g. status and switch on and off the circulation pump. Is there a list which status I can read out and can change for the fuel cell? It is very boring to try all possibilies and than run in the limitation of 1450.

By the way we can also talk on Germany, if you like.

woehrl01:
Thanks @SPITZGITHUB . I added a new PR with the new properties #204 .

PR #205, opened 2021-10-03
Add flat option to dump_secure

williwacker:
For having a short look on the dump output I find it more convenient having a flat output.

### DIFF
--- a/PyViCare/PyViCareDeviceConfig.py
+++ b/PyViCare/PyViCareDeviceConfig.py
@@ -80,6 +80,8 @@ class PyViCareDeviceConfig:
             return m.group(1) + ('#' * len(m.group(2))) + m.group(3)
 
         dumpJSON = re.sub(r'(["\/])(\d{6,})(["\/])', repl, dumpJSON)
-        if not flat:
-            return dumpJSON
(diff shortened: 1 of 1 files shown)


PyViCare/PyViCareDeviceConfig.py
```diff
@@ -73,10 +73,13 @@ def asAutoDetectDevice(self):
...
+        dumpJSON = json.dumps(self.get_raw_json(), indent=None if flat else 4)
 
         def repl(m):
             return m.group(1) + ('#' * len(m.group(2))) + m.group(3)
 
-        return re.sub(r'(["\/])(\d{6,})(["\/])', repl, dumpJSON)
+        dumpJSON = re.sub(r'(["\/])(\d{6,})(["\/])', repl, dumpJSON)
+        if not flat:
+            return dumpJSON
+        return re.sub(r'{"apiVersion', '\n{"apiVersion', dumpJSON)
```
woehrl01: There is no guarantee that "apiVersion" will be the first element in the object. You are also using a regex replace, in that case a normal string replace would be sufficient.

It would be great to have an alternative solution for adding newlines.

woehrl01 (changes requested):

Issue #193, opened 2021-09-26
Consumption data unit support


oischinger:
Turns out that some systems report kWh and others report m³.
Here's an example of m³ https://github.com/somm15/PyViCare/blob/master/tests/response/Vitodens200W.json ("value": "cubicMeter")
All the others seem to be in kWh ("value": "kilowattHour").

I wonder what is the best way to approach this:
a) Add an additional getter for the unit
b) Perform some calculations to always return kWh or m³

It would be best to understand Viessmann's strategy. Searching in their forum indicates that they are switching over to kWh because "that's what users asked for". Maybe they just didn't yet switch all systems to kWh? Also: It could be that not all systems are capable of delivering kWh since you cannot enter the "Brennwert", etc on the device.
If that's the case I would suggest we go with a) and perform some calculations in the HA Vicare integration to convert.

woehrl01:
I think having a both options could be handy. For the calculation part we need to pass the correct values for the gas.

oischinger:
Do we want to handle the calculation in HA or pyVicare?
Usually the HA Core devs want to offload as much as possible to the library.

In any way we need a configuration option for these values to do the calculation.

crazyfx1:
I think I would do the calculation in the HA integration, since it's easier to configure the "Brennwert" there if it's missing. But either way would be fine ;)

oischinger:
Me too. On the other hand HA devs tend to ask for offloading these kind of things to the library. So we might have to pass the Brennwert through

woehrl01:
I agree, I'm thinking about adding new options to the PyViCare class where you can generally set the output unit (and where you pass the appropriate convertion units. This results in having the a consistent configuration option on the root across all possible values. So it basically requires setting the values only once in the HA setup. Easily to pass via the config flow.

The reason for offloading it to the library is the there are multiple methods where this needs to be implemented. At I think this is way easier to manage and test of its implemented there.

crazyfx1:
Is this still needed? Seems like they support kWh now in the new beta. I'm just confused that it's just a frontend PR.
https://github.com/home-assistant/frontend/pull/10075

woehrl01:
I think we still need at least provide the unit via the api

oischinger:
Yes, sometimes the API only delivers cubic meters and we need to differentiate

PR #214, opened 2021-10-07
feat: Add consumption units for gas heatings

crazyfx1:
Adds consumption units for gas heatings as described in #193. If this is useful for HA, we can add the units for fuel cell aswell.


woehrl01 (approved):

oischinger:
Nice 👍

PR #218, opened 2021-10-11
Add units to fuel cell

crazyfx1:
Also add units to fuel cell. 

@woehrl01 Can you please create a new version after this commit?


woehrl01 (commented):
🚀

oischinger:
Does anyone know what are the possible values for the units? Cubic meters and kilowatt hours for gas consumption? But for power consumption? Only kWh or anything else?
@woehrl01 @crazyfx1 ?

woehrl01:
I haven't seen other units than the ones you mentioned in the wild.

PR #219, opened 2021-10-15
build: Add GitHub codespace support

woehrl01:
This PR adds support for GitHub Codespaces, where everything is set up to develop right on the spot.
* It also refactored the little scripts to a makefile to make the execution even more straightforward.
* Lastly, it changes the authentication in the integration test to the browser-based flow so that you only need to store the `PYVICARE_CLIENT_ID` in [your private Codespace secrets.](https://docs.github.com/en/codespaces/managing-your-codespaces/managing-encrypted-secrets-for-your-codespaces)


crazyfx1 (approved):
Nice 😁

PR #235, opened 2021-12-09
Vitodens-100W, new json dump, added new methods

damianbro:
I checked with new Vitodens-100W json dump, added new methods for gas/power consumption and added some simple tests for Vitodens-100W. I'm not sure if that should be done this way but IMO with new API methods its the only way.


woehrl01 (approved):

woehrl01:
This looks really good @damianbro. Thank you!

PR #241, opened 2022-01-09
Update README.md

poggenpower:
As it took me a while to figure out that the redirect URL need to be `vicare://oauth-callback/everest` and saw Users in die Discord channel running into the same issue. I think these would be an useful hints


crazyfx1:
There is already a description how to create the API keys with the correct redirect uri in the readme.md. Also unfortunately your description is full of typos and errors, I can't approve this.

crazyfx1:
I've moved the part in the current readme.md up so it's more visible for the user.

PR #267, opened 2022-09-20
Adding getBoilerTargetTemperature

mikissling:
Tested with a Vitdodens 200W B2HA, gives the boiler target temperature according to the current active heating curve. Viessmann API data point is heating.boiler.temperature

### DIFF
--- a/PyViCare/PyViCareGazBoiler.py
+++ b/PyViCare/PyViCareGazBoiler.py
@@ -95,6 +95,10 @@ class GazBoiler(Device):
         return self.service.getProperty("heating.boiler.sensors.temperature.main")["properties"]["value"]["value"]
 
     @handleNotSupported
+    def getBoilerTargetTemperature(self):
+        return self.service.getProperty("heating.boiler.temperature")["properties"]["value"]["value"]
+ 
+    @handleNotSupported
     def getDomesticHotWaterChargingLevel(self):
         return self.service.getProperty("heating.dhw.charging.level")["properties"]["value"]["value"]
 
--- a/tests/response/Vitodens200W_B2HA.json
+++ b/tests/response/Vitodens200W_B2HA.json
@@ -1,0 +1,2052 @@
+{
+    "data": [
+        {
+            "apiVersion": 1,
+            "commands": {},
+            "deviceId": "0",
+            "feature": "heating.circuits.2.frostprotection",
+            "gatewayId": "################",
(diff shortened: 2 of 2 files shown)


crazyfx1:
Thanks for your contribution.
Can you please add unit tests for the new property by using your response file?

mikissling:
I am happy to do - but I think I need a bit of guidance. Where do I add the unit test? As you may guess I am just starting to get my head around PyViCare, Python and github.

crazyfx1:
No problem. 
If you check the folder `tests` in the repository, you'll find unit tests for all kinds of heating systems. 
If would be the easiest if you copy an existing unit test file (for example `test_Vitodens_100W.py`) and change it to use your new response file and adapt the test cases.

mikissling:
I must admit I do not understand the concept and workflow of the unit test fully but I anyway tried to start somwhere:

- As proposed I took the test_Vitodens_100W.py and renamed/adapted it to use my response file.
- So far I have included only a test for the newly introduced function getBoilerTargetTemperature() - or is it meant to have a function for all values that my heating returns in the response file? I see the some of the other test files are longer, some shorter.
- What is the expected output of the script? I did not manage to run it so far. When I run the script here on my Windows computer I get the following output - I guess it is related to my folder structure?! 

```
====== RESTART: C:\python\PyViCare-master\tests\test_Vitodens200W_B2HA.py ======
Traceback (most recent call last):
File "C:\python\PyViCare-master\tests\test_Vitodens200W_B2HA.py", line 4, in <module> 
from tests.ViCareServiceMock import ViCareServiceMock
ModuleNotFoundError: No module named 'tests'
```

Before I commit anything this is how it currently looks:

```
import unittest

from PyViCare.PyViCareGazBoiler import GazBoiler
from tests.ViCareServiceMock import ViCareServiceMock


class Vitodens200W_B2HA(unittest.TestCase):
    def setUp(self):
        self.service = ViCareServiceMock('response/Vitodens200W_B2HA.json')
        self.device = GazBoiler(self.service)

    def test_getBoilerTargetTemperature(self):
        self.assertEqual(self.device.getBoilerTargetTemperature(self), 5)
```

crazyfx1:
The test files are not meant to be run. They are test definitions for the pytest test-runner. You can run them by running the pytest command from console in the root folder (you might need to install it first with pip). 

The test looks fine, thanks!

PR #275, opened 2022-11-17
Add more fuel cell functions

nagelp:
I've now added most Fuel Cell data points as getFuelCell... methods. However they're all only available in the paid "Advanced" API plan - currently I've only marked them as such with a single comment. New users who don't see that comment might be confused why get errors when they call these methods, even though others work...


woehrl01:
@nagelp Thank you very much for the PR. No worries, it's pretty common that some methods don't return values for some users. Can you please add some tests for your added methods, so we can confirm that they work? Especially after Viessmann again changes the API?

Please also remove the TODO comments. Thanks!

nagelp:
> @nagelp Thank you very much for the PR. No worries, it's pretty common that some methods don't return values for some users. Can you please add some tests for your added methods, so we can confirm that they work? Especially after Viessmann again changes the API?
> 
> Please also remove the TODO comments. Thanks!

Done.

PR #282, opened 2022-11-27
feat: allow setting sub digit temperature

### LINKED ISSUE #280: Feature: Enable 0.5 steps on setTemperature
https://github.com/somm15/PyViCare/issues/279#issuecomment-1321200568

some devices allow setting the temperature in 0.5 steps instead of full numbers.

woehrl01:
Fixes #280 

@crazyfx1 I'm a bit concerned about the implications of this. If other integrations are just passing in a float `21.5` in the past it got converted to `21` which every API could handle. If it now passes `21.5` some integrations could now fail, as the API does not allow it. What do you think? Should we take the effort to cast it to the right fraction?

### DIFF
--- a/PyViCare/PyViCareDevice.py
+++ b/PyViCare/PyViCareDevice.py
@@ -45,7 +45,7 @@ class Device:
     """This class connects to the Viesmann ViCare API.
     The authentication is done through OAuth2.
     Note that currently, a new token is generate for each run.
-    """
+    """ 
 
     def __init__(self, service: ViCareService) -> None:
         self.service = service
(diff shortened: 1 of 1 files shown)


oischinger:
@woehrl01 did you find any indication of which devices supports settings 0.5 steps? Looks like this information is not available on the API, right?

woehrl01:
It's available on the API level, at least it's in the dump. Unfortunately, I don't receive any data currently from the API for my device. App works fine.

@oischinger

oischinger:
are you saying that every Viessmann device supports 0.5 steps?

woehrl01:
No, that depends on the device.

PR #288, opened 2022-12-17
Added a timeout for requests.get()

nagelp:
(31 seconds hard-coded for now) - this should avoid hanging in case network or server go down during a request.

Should we handle the exception and do a retry within PyViCare, or leave that up to the user of PyViCare? I didn't add any handling of TimeoutError, urllib3.exceptions.ReadTimeoutError or requests.exceptions.ReadTimeout for now.

### DIFF
--- a/PyViCare/PyViCareAbstractOAuthManager.py
+++ b/PyViCare/PyViCareAbstractOAuthManager.py
@@ -34,7 +34,7 @@ class AbstractViCareOAuthManager:
     def get(self, url: str) -> Any:
         try:
             logger.debug(self.__oauth)
-            response = self.__oauth.get(f"{API_BASE_URL}{url}").json()
+            response = self.__oauth.get(f"{API_BASE_URL}{url}", timeout=31).json()
             logger.debug(f"Response to get request: {response}")
             self.__handle_expired_token(response)
             self.__handle_rate_limit(response)


woehrl01:
@nagelp 
Thank you! I don't think a retry is needed in that case and would leave it up to the user for now. Out of curiosity: Why did you choose 31 and not a flat 30 seconds timeout?

nagelp:
> Out of curiosity: Why did you choose 31 and not a flat 30 seconds timeout?

I followed the recommendation from https://requests.readthedocs.io/en/latest/user/advanced/#timeouts - "It’s a good practice to set connect timeouts to slightly larger than a multiple of 3, which is the default [TCP packet retransmission window](https://www.hjp.at/doc/rfc/rfc2988.txt)."
I don't think it really matters, and I'm not even sure this recommendation makes much sense, because the TCP retransmission window is variable AFAIK. But I thought they must have had their reasons and just picked the slightly unaesthetic 31 over 30 ;)
Bigger question is whether the Viessmann server has been known to be really slow at times? Might there be a case where it won't answer within 31s?

woehrl01:
I see, thank you!

I'm not aware of a "common" slow response. But if its longer than 30 seconds, I'd argue I would still throw a timeout.

woehrl01 (approved):

Issue #283, opened 2022-11-29
Add data point "volumetricFlow" from Viessmann API


denmaus:
In the Basic version of the API there is a data point `heating.sensors.volumetricFlow.allengra`.
This data point was added on august 2022 into the api see [link](https://www.viessmann-community.com/t5/The-Viessmann-API/viessmannapi-2-0-9-und-IOBroker-Volumenstrom/m-p/250727/highlight/true#M866).

Can this data point be added to the library?

I have a Vitodens 343-F heater which is integrated via home assistant and I would like to be able to access this value there.

Screenshot from API documentation:
![image](https://user-images.githubusercontent.com/28492537/204560215-62419ac9-53ac-4703-8cc3-9120da63b833.png)

In the logging of the vicare integration this data point is listed as follows:

```json
        {
          "apiVersion": 1,
          "commands": {},
          "deviceId": "0",
          "feature": "heating.sensors.volumetricFlow.allengra",
          "gatewayId": "################",
          "isEnabled": true,
          "isReady": true,
          "properties": {
            "status": {
              "type": "string",
              "value": "connected"
            },
            "value": {
              "type": "number",
              "unit": "liter",
              "value": 312
            }
          },
          "timestamp": "2022-11-29T15:37:35.344Z",
          "uri": "https://api.viessmann.com/iot/v1/equipment/installations/######/gateways/################/devices/0/features/heating.sensors.volumetricFlow.allengra"
        },
```

crazyfx1:
Can you please provide a full API dump of your device?

denmaus:
I created a diagnosis dump from home assistant and removed the integration specific parts from the file.
My heating device is Vitodens 343-F B3UF with solar support for domestic hot water.

[vicare-json.txt](https://github.com/somm15/PyViCare/files/10203368/vicare-json.txt)

denmaus:
I created a dump from my device according to https://github.com/somm15/PyViCare#more-different-devices-for-test-cases-needed and started a pull requst (https://github.com/somm15/PyViCare/pull/295).

I hope it helps.

denmaus:
I see the measurement of interest `heating.sensors.volumetricFlow.allengra` is already available for FuelCell device here: https://github.com/somm15/PyViCare/blob/23efc1c3b551704be1995a81c0f69485df88bff8/PyViCare/PyViCareFuelCell.py#L133-L135

How to make it available also for Gas Boiler?

woehrl01:
[...]

PR #296, opened 2023-01-16
feat: add burner volumetricFlow sensor to gas devices

denmaus:
Add sensor for volumetric flow on the return.

Measurement can also be viewed in the menu of device via "Information > Burner > Flow sensor".

See also https://github.com/somm15/PyViCare/issues/283.


woehrl01:
Available in `2.24.0`

PR #307, opened 2023-04-07
preparation for multi device type support

simon-schuster:
Short explanation on what i did here:

- `PyViCareDevice` is now a base class for all different types of devices
- `PyViCareHeatingDevice` contains most of the functionality previously contained by the `PyViCareDevice`
- All types of heating devices now inherit from `PyViCareHeatingDevice`
- Sensors and radiators just inherit from `PyViCareDevice`
- `PyViCareGenericDevice` is the new generic device which has to inherit all types of devices which will be added
- added directories for better overview of the different device types


woehrl01:
Hi @simon-schuster,

Thank you very much for your pull request. I'm not feeling that comfortable with the changes. If we change the root device name and introducing folders, we will introduce a breaking change for everyone, where I don't see a big benefit. I prefer just introducing a new class for heating devices and keep the structure as is. Please also remove the new generic device, and simply use the new heating device.

simon-schuster:
Hi @woehrl01,

Thanks for taking the time to review.
So you mean just copy the current device code and call it heating device, but leaving the current device as it is?
Wouldn't that be a lot duplicated code and also maybe hard to understand where to add new functions in the future?

woehrl01:
No, the idea is to just add the HeatingDevices, move the stuff from Device to the new HeatingDevice, and use the HeatingDevice as the generic device.

simon-schuster:
@woehrl01 I have now reverted the folder structure to how it was before and move everything from device to heating device.
Should I also revert the recursive file search in `tests/test_TestForMissingProperties.py`or is it ok to leave it as it is for possible folders in the future?

tests/test_TestForMissingProperties.py
```diff
@@ -116,18 +116,30 @@ def find_feature_in_code(self, all_python_files, feature):
...
         all_python_files = {}
 
         for python in python_files:
-            if not python.endswith(".py"):
+            if not python[1].endswith(".py"):
                 continue
 
-            with open(join(python_path, python)) as f:
-                all_python_files[python] = f.read()
+            with open(join(python[0], python[1])) as f:
```
woehrl01: What are index 0 and index 1? Please make this more readable by renaming that

tests/test_TestForMissingProperties.py
```diff
@@ -116,18 +116,30 @@ def find_feature_in_code(self, all_python_files, feature):
...
+                all_python_files[python[1]] = f.read()
+
         return all_python_files
 
+    def get_all_files(self, files, path):
```
woehrl01: Please return a new array, instead of passing them as an argument. Or at least provide a second method which hides this and rename this method to include _recursive the more more self explanatory.

tests/test_TestForMissingProperties.py
```diff
@@ -116,18 +116,30 @@ def find_feature_in_code(self, all_python_files, feature):
 
     def read_all_python_code(self):
         python_path = join(dirname(__file__), '../PyViCare')
-        python_files = [f for f in listdir(python_path) if isfile(join(python_path, f))]
+        # stores pairs of [path, filename]
```
woehrl01: Please create a simple object to hold this.

PyViCare/PyViCareRoomSensor.py
```diff
@@ -1,8 +1,8 @@
-from PyViCare.PyViCareDevice import Device
+from PyViCare.PyViCareHeatingDevice import HeatingDevice
 from PyViCare.PyViCareUtils import handleNotSupported
 
 
-class RoomSensor(Device):
+class RoomSensor(HeatingDevice):
```
woehrl01: Those should inherit from a device, like you had before.

woehrl01 (changes requested):

woehrl01:
@simon-schuster thank you. I left some comments, you can leave the changes if you make some additional ones 😊

PR #314, opened 2023-05-13
handle InvalidTokenError and fix setProperty

williwacker:
After installing the latest release I discovered that after some time I always get an InvalidTokenError.
And I no longer can set any property


crazyfx1:
Hi Werner!
Thanks for your contribution. Could you please check the tests, the tested url needs to be adjusted.

williwacker:
Have updated the test as well. Anything else I have to do?
Don't know how to trigger a new action in GitHub.
Thanks

crazyfx1:
Thanks!

PR #318, opened 2023-05-28
Add Support for HEMS, VX3 and other EES Devices

Nibot1:
I would like to add Data Point Support for:
- HEMS (Energy Management)
- Vitocharge VX3 (Inverter and Battery Storage)

Data points that would be supported: ess.* photovoltaic.* pcc.*

Sadly, to use these Data Points, you need the Advanced or Electric Subscription.

### DIFF
--- a/PyViCare/PyViCareDeviceConfig.py
+++ b/PyViCare/PyViCareDeviceConfig.py
@@ -11,6 +11,7 @@ from PyViCare.PyViCareOilBoiler import OilBoiler
 from PyViCare.PyViCarePelletsBoiler import PelletsBoiler
 from PyViCare.PyViCareRadiatorActuator import RadiatorActuator
 from PyViCare.PyViCareRoomSensor import RoomSensor
+from PyViCare.PyViCareElectricalEnergySystem import ElectricalEnergySystem
 
 logger = logging.getLogger('ViCare')
 logger.addHandler(logging.NullHandler())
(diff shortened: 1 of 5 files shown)


crazyfx1:
Thank you very much!

PyViCare/PyViCareDeviceConfig.py
```diff
@@ -71,7 +75,11 @@ def asAutoDetectDevice(self):
             (self.asOilBoiler, r"Vitoladens|Vitoradial|Vitorondens|VPlusH|V200KW2_6", []),
             (self.asPelletsBoiler, r"Vitoligno|Ecotronic|VBC550P", []),
             (self.asRadiatorActuator, r"E3_RadiatorActuator", ["type:radiator"]),
-            (self.asRoomSensor, r"E3_RoomSensor", ["type:climateSensor"])
+            (self.asRoomSensor, r"E3_RoomSensor", ["type:climateSensor"]),
+            (self.asElectricalEnergySystem, r"E3_HEMS", ["type:hems"]),
+            (self.asElectricalEnergySystem, r"E3_TCU10_x07", ["type:tcu"]),
```
CFenner, line 80: @Nibot1, why did you choose to handle the `tcu` as an *ElectricalEnergySystem*? Due to it's `device id` `gateway` which is set to `0` in `PyViCare.py` this leads to duplicating device `0`. Instead this should be handled as `gateway` (recently added class) to expose its feature (currently only wifi strength). What do you think?

Nibot1: Hi, apologies for my delayed response. You are right, we should handle the TCU as a Gateway. At the time when I implemented this, I didn't exactly know what the meaning of TCU was. Now I think TCU is Transmission Control Unit, is this right?

CFenner: I think so too, but I don't have the device in my setup.

PR #319, opened 2023-06-08
Add ability to set DHW operating mode 

pavelsidla:
Opearting mode set ability to (eco,comfort,off)

### DIFF
--- a/PyViCare/PyViCareHeatingDevice.py
+++ b/PyViCare/PyViCareHeatingDevice.py
@@ -196,6 +196,11 @@ class HeatingDevice:
         return self.service.setProperty("heating.dhw.temperature.temp2", "setTargetTemperature",
                                         {"temperature": int(temperature)})
 
+    @handleAPICommandErrors
+    def setDhwOperatingMode(self, mode):
+        return self.service.setProperty("heating.dhw.operating.modes.active", "setMode",
+                                        {'mode': mode})
+
     @handleNotSupported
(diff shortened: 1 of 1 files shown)


pavelsidla:
Hello can you add this feature, i need to set different operating mode in day schedule.
FG: 
I am going to bath around 18:00-22:00, so i need to change operating mode to comfort because i wont to wait till water will be heated.

PyViCare/PyViCareHeatingDevice.py
```diff
@@ -196,6 +196,11 @@ def setDomesticHotWaterTemperature2(self, temperature):
         return self.service.setProperty("heating.dhw.temperature.temp2", "setTargetTemperature",
                                         {"temperature": int(temperature)})
 
+    @handleAPICommandErrors
+    def setDhwOperatingMode(self, mode):
```
crazyfx1:
```suggestion
    def setDomesticHotWaterOperatingMode(self, mode):
```

crazyfx1: Could you please change the method name to be consistent with the other method names?

crazyfx1 (changes requested):

PyViCare/PyViCareHeatingDevice.py
pavelsidla: Hi its done :)

crazyfx1 (approved):

PR #334, opened 2023-10-19
feat: add PyViCareGateway class

### LINKED ISSUE #330: Read WiFi Signal Strength
Hey,

I've been looking into the code and into the HomeAssistant integration lately. I would be interested in reading gateway specific values with the API and I found `device.wifi`. But this seems not to be in the library code yet. Where could this be implemented? Is there a dedicated Gateway class necessary for it?

CFenner:
fixes #330 
fixes the issue that gateways are handled as generic devices / heating devices.

Related to https://github.com/somm15/PyViCare/issues/335


CFenner:
Can someone have a look at this?

woehrl01:
@CFenner thank you for your pr. Can you please write a unit test, backed by some actual data from your device? Thank you!

CFenner:
Unfortunately, I have issues setting up my local dev environment to run proper tests. I doubt that this works, as the gateway also has a different endpoint to fetch features.

CFenner:
Would be good to at least have also a response from an OPTO2.

CFenner:
@woehrl01 please have another look.

PyViCare/PyViCareService.py
```diff
@@ -61,4 +64,6 @@ def setProperty(self, property_name: str, action: str, data: Any) -> Any:
 
     def fetch_all_features(self) -> Any:
         url = f'/features/installations/{self.accessor.id}/gateways/{self.accessor.serial}/devices/{self.accessor.device_id}/features/'
+        if self._isGateway():
+            url = f'/features/installations/{self.accessor.id}/gateways/{self.accessor.serial}/features/'
```
woehrl01, line 69: Can you please add a test that this url gets called for your Gateway device?

woehrl01 (changes requested):
Looks great, one last test please. Thank you!

There should be already a similar test, which accepts the roles you are using for detection of the right call: test_PyVivareService

PyViCare/PyViCareService.py
CFenner: Is there any sample I can reuse?

CFenner: Any suggestion how to test this?

woehrl01: Just copy this function and set the roles on the service like you did on your other test

https://github.com/somm15/PyViCare/blob/16eb10d1c5fdb372a669705b54e34bb222b2057b/tests/test_PyViCareService.py#L14

tests/test_PyViCareDeviceConfig.py
```diff
@@ -71,3 +71,9 @@ def test_autoDetect_VitoconnectOpto1_asGateway(self):
         c = PyViCareDeviceConfig(self.service, "0", "Heatbox1", "Online")
         device_type = c.asAutoDetectDevice()
         self.assertEqual("Gateway", type(device_type).__name__)
+
+    def test_autoDetect_RoleGateway_asGateway(self):
+        self.service.hasRoles = has_roles(["type:gateway;VitoconnectOpto1"])
```
woehrl01, line 76: This one ☝️😊

woehrl01:
@CFenner sorry for missleading you on the shown test, I thought you have modified a different method and not the `fetch_all_features` one. Do you feel comfortable to adapt the test to that function? (you can force-push override my changes)

CFenner:
> I thought you have modified a different method and not the `fetch_all_features` one. 

I probably need to change both, right?

tests/test_PyViCareService.py
```diff
@@ -2,26 +2,35 @@
...
     def test_setProperty_string(self):
-        self.service.setProperty("someprop", "doaction", '{}')
+        service = ViCareService(self.oauth_mock, self.accessor, [])
+        service.setProperty("someprop", "doaction", '{}')
         self.oauth_mock.post.assert_called_once_with(
             '/features/installations/[id]/gateways/[serial]/devices/[device]/features/someprop/commands/doaction', '{}')
+
+    def test_getProperty_gateway(self):
+        service = ViCareService(self.oauth_mock, self.accessor, ["type:gateway;VitoconnectOpto1"])
+        service.getProperty("someprop")
```
woehrl01: replacing this call with `fetch_all_features` should be enough

woehrl01:
@CFenner yes, both methods needs to be changed to point to the different url, also please add a test for the original `fetch_all_features` value, too. Looks like this was missing. Thank you!

PyViCare/PyViCareService.py
```diff
@@ -45,13 +41,21 @@ def __init__(self, oauth_manager: AbstractViCareOAuthManager, accessor: ViCareDe
...
+        if self._isGateway():
+            return f'/features/installations/{self.accessor.id}/gateways/{self.accessor.serial}/features/{property_name}'
+        return f'/features/installations/{self.accessor.id}/gateways/{self.accessor.serial}/devices/{self.accessor.device_id}/features/{property_name}'
+
 
     def hasRoles(self, requested_roles) -> bool:
         return hasRoles(requested_roles, self.roles)
 
+    def _isGateway(self) -> bool:
+        return self.hasRoles(["type:gateway;VitoconnectOpto1"])
```
CFenner, line 57: Is it fine to just check the roles here and not the model as in `asAutoDetectDevice`?

woehrl01: yes, I would keep that until, we have a second device which does not work correctly

PyViCare/PyViCareDeviceConfig.py
```diff
@@ -54,6 +55,9 @@ def asRoomSensor(self):
     
     def asElectricalEnergySystem(self):
         return ElectricalEnergySystem(self.service)
+
```
CFenner, line 58:
```suggestion

```

PyViCare/PyViCareService.py
```diff
@@ -45,13 +41,21 @@ def __init__(self, oauth_manager: AbstractViCareOAuthManager, accessor: ViCareDe
         self.roles = roles
 
     def getProperty(self, property_name: str) -> Any:
-        url = buildGetPropertyUrl(
-            self.accessor, property_name)
+        url = self.buildGetPropertyUrl(property_name)
         return self.oauth_manager.get(url)
+
```
CFenner, line 46:
```suggestion

```

tests/response/VitoconnectOpto1.json
```diff
@@ -0,0 +1,65 @@
...
+                    "type": "number",
+                    "unit": "",
+                    "value": -69
+                }
+            },
+            "timestamp": "2023-12-26T20:44:41.417Z",
+            "uri": "https://api.viessmann.com/iot/v1/features/installations/#######/gateways/################/features/gateway.wifi"
+        }
+    ]
+}
```
CFenner, line 65:
```suggestion
}

```

woehrl01:
awesome, thank you for this addition @CFenner

woehrl01:
awesome, thank you for this addition @CFenner

PyViCare/PyViCareService.py
CFenner: Probably `Heatbox2_SRC` and `E3_TCU10_x07`/`type:gateway;TCU300` should be handled as Gateways as well.

PR #345, opened 2023-12-21
chore(testdata): Add Vitocal 333-G data

edeckert:


CFenner:
@edeckert I'm working on another PR #360 to detect if a device is a heating, dhw, solar or ventilation device. When looking at your sample data I notice that you have no `"feature": "heating"` in your response. Is this a pure DHW device?

PR #346, opened 2023-12-22
docs: update Viessmann API client description

CFenner:
Looks like the developer portal changed a bit. I updated the description accordingly.


woehrl01:
Thanks!

PR #354, opened 2024-01-21
Add getSerial for gateways

CFenner:


tests/test_VitoconnectOpto1.py
```diff
@@ -9,6 +9,10 @@ def setUp(self):
         self.service = ViCareServiceMock('response/VitoconnectOpto1.json')
         self.device = Gateway(self.service)
 
+    def test_getSerial(self):
+        self.assertEqual(
+            self.device.getSerial(), "#########")
```
CFenner:
```suggestion
            self.device.getSerial(), "################")
```

PyViCare/PyViCareGateway.py
```diff
@@ -4,6 +4,10 @@
 
 class Gateway(Device):
 
+    @handleNotSupported
+    def getSerial(self):
+        return self.service.getProperty("gateway.devices")["gatewayId"]
```
CFenner, line 9: At least my Heatbox1 / VitoconnectOpto1 has no device.serial data point. I requested it in the [Viessmann community](https://community.viessmann.de/t5/Feedback-API/Serial-Number-an-Gateways/td-p/410253) but no reply so far. So I get the serial from another data point.

PR #356, opened 2024-01-21
Make HeatingDevice inherit from Device

CFenner:
This makes `Device` the common base class also for `HeatingDevice`s.

PR #360, opened 2024-02-09
feat: Expose if a device supports dhw, solar thermal or ventilation capabilities

CFenner:
This adds methods to check if a device supports general heating, dhw, solar thermal or vertilation capabilities.

This requires #356

Bildschirmfoto 2024-04-05 um 08 29 23


CFenner:
Can we bring this in, it would be essential to support ventilation devices in HA.

PyViCare/PyViCareDevice.py
```diff
@@ -15,3 +15,19 @@ def __init__(self, service: ViCareService) -> None:
     @handleNotSupported
     def getSerial(self):
         return self.service.getProperty("device.serial")["properties"]["value"]["value"]
+
+    @handleNotSupported
+    def isHeatingDevice(self):
```
Nibot1: What do you think of renaming this function in something like `isDeviceHeatingEnabled` ? Because I don't know which impact on this data point is made by changing the active mode of the Heating Device (e.g. DHW only during Summer)
I think the same renaming could be useful for the other functions as well.

PyViCare/PyViCareDevice.py
```diff
@@ -15,3 +15,19 @@ def __init__(self, service: ViCareService) -> None:
...
+        return self.service.getProperty("heating")["isEnabled"]
```
Nibot1: Would it be nice to catch the `PyVicareNotSupported` error internally, rather than relying on the implementation of error handling from the Users?

woehrl01: Agree, this should return false if it does not exists.

PyViCare/PyViCareDevice.py
CFenner: Do you have a system where you could verify this?

Nibot1: I can test it later with my Oil Boiler and Heatbox2

CFenner: There is also a status

Bildschirmfoto 2024-04-03 um 11 13 50

tests/test_Vitovent.py
```diff
@@ -0,0 +1,22 @@
...
+class Vitovent(unittest.TestCase):
+    def setUp(self):
+        self.service = ViCareServiceMock('response/Vitovent.json')
+        self.device = VentilationDevice(self.service)
+
+    def test_isHeatingDevice(self):
+        self.assertFalse(self.device.isHeatingDevice())
+
+    def test_isDomesticHotWaterDevice(self):
+        self.assertTrue(self.device.isDomesticHotWaterDevice())
```
CFenner: For some reason the vitovent says it is a dhw device. Maybe this is all not that accurate.

PyViCare/PyViCareDevice.py
Nibot1: Yes you are right. I have tested it, changing the mode to DHW Only doesn't affect the ´isEnabled` flag for heating on my System.

CFenner:
@woehrl01 what shall I do about the failing test for newly added properties?

woehrl01:
@CFenner either implement the methods or add them to the ignore list (I'm fine with the later)

PyViCare/PyViCareDevice.py
CFenner: Can you also verify if this influences the ["properties"]["active"]["value"] of DHW? 
So does turning of the DHW (is that possible?) disable the flag?

tests/test_Vitovent.py
CFenner: @SHuechtemann can you say why your Vitovent has an active DomesticHotWater flag?

CFenner:
There seem to be cases where `"feature": "heating"` is not available at all: https://github.com/somm15/PyViCare/pull/327 https://github.com/somm15/PyViCare/pull/345

PyViCare/PyViCareDevice.py
Nibot1: If I disable DHW on my boiler, the feature `heating.dhw` changes `isEnabled` to false and removes the properties
`{
            "feature": "heating.dhw",
            "gatewayId": "xxxx",
            "deviceId": "0",
            "timestamp": "2024-04-04T16:01:14.844Z",
            "isEnabled": false,
            "isReady": true,
            "apiVersion": 1,
            "uri": "https://api.viessmann.com/iot/v1/features/installations/xxx/gateways/xxxxxx/devices/0/features/heating.dhw",
            "properties": {},
            "commands": {}
        }`
To disable dhw I need to reconfigure my boiler and remove the dhw temp. Sensor
So my Boiler acts as a device where DHW is possible but not enabled (probably `isReady` shows this)

Nibot1:
@CFenner my boiler also doesn't show up the `"feature": "heating"` what about testing the `"feature": "heating.circuits"` to solve this? I would believe that a device which is capable of heating must show a Circuit, and the devices which don't have heating capabilities won't show circuits.

CFenner:
Yes, I can check that. 

I wonder if this could also be due to outdated api dumps.

Nibot1:
I don't think so because my API dumps are just a few minutes old.

tests/test_Vitovent.py
SHuechtemann: Idk. Maybe the description is inacuarate/misleading as my vivovent has an active Heat exchanger and is capable to heat up the incoming air...

SHuechtemann: Also it has a bypass function to bypass the heat exchanger. Maybe this is what the flag is for, showing if bypass is active or inactive. This should be in the Response from the API, as the viessmann App also shows this.

LarsTh86:
Is there progress for the implementation of the ventilation? can I test something?

LarsTh86:
How i can test this?

PR #364, opened 2024-03-30
Remove HEMS an EEBus devices and change TCU handling to gateway

Nibot1:
This PR will remove the handling of HEMS and EEBUS devices because the API does not provide any features or data points.
The Functionality of managing the Vitocharge VX3 is not affected, the HEMS and EEBus devices are separate devices, which I accidentally messed up in my last PR
The handling of the E3_TCU unit will be changed to a gateway, like mentioned by @CFenner in PR #318
Maybe in the future there will be more possibilities to manage, for example Vehicle Charging Stations (VCS).
The devices are already listed, but we cannot access features or data points.

Sadly, I cannot test the TCU data point for Wi-Fi strength because my TCU is connected via Ethernet. There is a documented data point for Ethernet `gateway.ethernet.config`, but this also somehow doesn't show up on my TCU, so I can't implement this currently.

### DIFF
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -53,7 +53,7 @@ class PyViCare:
             for gateway in installation.gateways:
                 for device in gateway.devices:
                     if device.deviceType not in ["heating", "zigbee", "vitoconnect", "electricityStorage", "tcu", "ventilation"]:
-                        continue  # we are only interested in heating, photovoltaic, electricityStorage, hems and ventilation devices
+                        continue  # we are only interested in heating, photovoltaic, electricityStorage, and ventilation devices
 
                     accessor = ViCareDeviceAccessor(
                         installation.id, gateway.serial, device.id)
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -55,9 +55,6 @@ class PyViCare:
                     if device.deviceType not in ["heating", "zigbee", "vitoconnect", "electricityStorage", "tcu", "ventilation"]:
                         continue  # we are only interested in heating, photovoltaic, electricityStorage, hems and ventilation devices
 
-                    if device.id == "gateway" and device.deviceType == "vitoconnect":
-                        device.id = "0"  # vitoconnect has no device id, so we use 0
-
                     accessor = ViCareDeviceAccessor(
                         installation.id, gateway.serial, device.id)
                     service = self.__buildService(accessor, device.roles)
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -52,20 +52,11 @@ class PyViCare:
         for installation in self.installations:
             for gateway in installation.gateways:
                 for device in gateway.devices:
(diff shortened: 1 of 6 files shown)


PyViCare/PyViCare.py
```diff
@@ -52,21 +52,12 @@ def __extract_devices(self):
         for installation in self.installations:
             for gateway in installation.gateways:
                 for device in gateway.devices:
-                    if device.deviceType not in ["heating", "zigbee", "vitoconnect", "electricityStorage", "EEBus", "hems", "tcu", "ventilation"]:
+                    if device.deviceType not in ["heating", "zigbee", "vitoconnect", "electricityStorage", "tcu", "ventilation"]:
                         continue  # we are only interested in heating, photovoltaic, electricityStorage, hems and ventilation devices
 
                     if device.id == "gateway" and device.deviceType == "vitoconnect":
```
CFenner: Probably this is also no longer necessary.

CFenner (commented):
Can you also add test cases as for the OPTO1?

https://github.com/somm15/PyViCare/blob/master/tests/test_VitoconnectOpto1.py

PyViCare/PyViCare.py
Nibot1: Yes you are right, this is not necessary anymore.

Nibot1:
Sadly, I cannot test the TCU data point for Wi-Fi strength because my TCU is connected via Ethernet, but I will write a test for the Serial number

CFenner (approved):
👍 

You could also add auto detect test cases: https://github.com/somm15/PyViCare/blob/master/tests/test_PyViCareDeviceConfig.py

PyViCare/PyViCareDeviceConfig.py
```diff
@@ -84,10 +84,8 @@ def asAutoDetectDevice(self):
             (self.asPelletsBoiler, r"Vitoligno|Ecotronic|VBC550P", []),
             (self.asRadiatorActuator, r"E3_RadiatorActuator", ["type:radiator"]),
             (self.asRoomSensor, r"E3_RoomSensor", ["type:climateSensor"]),
-            (self.asElectricalEnergySystem, r"E3_HEMS", ["type:hems"]),
-            (self.asElectricalEnergySystem, r"E3_TCU10_x07", ["type:tcu"]),
-            (self.asElectricalEnergySystem, r"E3_EEBus", ["type:eebus"]),
-            (self.asElectricalEnergySystem, r"E3_VitoCharge_03", ["type:energy_storage"]),
+            (self.asGateway, r"E3_TCU10_x07", ["type:gateway;TCU300"]),
```
CFenner, line 87: The TCU could probably also be ignored in the HA integration, right?

https://github.com/home-assistant/core/blob/dev/homeassistant/components/vicare/const.py#L18

Nibot1: Yes, the TCU does not provide any information

Nibot1: Do you create the PR in the Home Assistant Repo?

woehrl01:
Thank you @Nibot1 lgtm

PR #366, opened 2024-04-04
chore(testdata): Add Vitoladens 300-C data

Nibot1:


CFenner:
sample data added with #616

PR #371, opened 2024-06-26
feat: Add methods for heat pump cooling consumption

CFenner:
Adds methods to access the power consumption used for cooling on heat pumps.

This is actually not documents in the [DataPoints](https://documentation.viessmann.com/static/iot/data-points)(yet?), but it is in the latest API responses for a heat pump and in the ViCare app.

342651323-2812795a-6499-4112-a2d8-34725920df1f

See also 
https://github.com/home-assistant/core/issues/120396#issuecomment-2188361639

fyi @x4N70pHyLL


CFenner:
@x4N70pHyLL is this accessible via the basic API?

x4N70pHyLL:
@CFenner Yes both the screenshot from the App and the Diag Info is from the free plan, I don't have any premium stuff.

Issue #402, opened 2024-09-23
Datapoint not included in HA


tpulatha:
Hi,

I have a Vitovalor PT2 Fuelcell/boiler/hot water cylinder combination which I have connected to Home Assistant since quite a while. Unfortunately, It seems to be missing quite some datapoints. For example `heating.dhw.sensors.temperature.hotWaterStorage.bottom`

When checking the debuglogs, I do see the information

```
            "deviceId": "0",
            "feature": "heating.dhw.sensors.temperature.hotWaterStorage.bottom",
            "gatewayId": "################",
            "isEnabled": true,
            "isReady": true,
            "properties": {
              "status": {
                "type": "string",
                "value": "connected"
              },
              "value": {
                "type": "number",
                "unit": "celsius",
                "value": 50.6
              }
            },
```

I already deleted and readded the device in the integration. Any other idea? Of course latest HA version 2024.9.2. Thanks!

PS: I also posted this in your fork, sorry, didn't know which one should be used.

CFenner:
Hi @tpulatha 

Just checked for your data point here in the library, it was removed recently (see https://github.com/openviess/PyViCare/pull/381, you should also see a hint for that in your diagnostic information) but it's replacement  `heating.dhw.sensors.temperature.dhwCylinder.bottom` is exposed via `getHotWaterStorageTemperatureBottom`. So no issue from the library pov.

But there is no sensor in the HA integration that uses this value. Therefore I ask you to open an issue in https://github.com/home-assistant/core

PS: Currently the library for is used in the HA integration, but we plan to switch it back soon.

tpulatha:
Thanks, yeah sorry I realized that it was deprecated after I posted this. Ok, I'll open an issue. I try to add all missing sensors.

CFenner:
We have `dhwCylinder`, `dhwCylinder.top`and `dhwCylinder.bottom` exposed here. So at least `dhwCylinder.middle` and `dhwCylinder.midBottom` are missing.

CFenner:
@tpulatha can you add a diagnostics file, please?

tpulatha:
Added below. Sorry for the delay. 

[config_entry-vicare-01J7Y6PA4RYJY92JQDB3EER28G.json](https://github.com/user-attachments/files/17155919/config_entry-vicare-01J7Y6PA4RYJY92JQDB3EER28G.json)

CFenner:
@tpulatha this should be in production by now.. can you verify and close the issue?

[...]

PR #404, opened 2024-09-25
feat: add new sensors for DomesticHotWaterStorage

CFenner:
related to #402


CFenner:
@tpulatha please provide the data of your device, so I can add some tests.

tpulatha:
Added in the issue #402.

PR #484, opened 2025-01-18
added version for updated 050W

joearkay:
I noticed when using the [Home Assistant integration](https://www.home-assistant.io/integrations/vicare/) that relies on this library that some features were not showing up as entities. I think this is due to some deprecated APIs which are present in the current 050W configuration.

For me, I want to use the `"feature": "heating.dhw.operating.modes.active"` to set between comfort/eco/off

Please let me know if you would prefer me to update the Vitodens050W.json instead of creating a _2, however I didn't want to introduce any breaking changes. 

I haven't yet had the time to study the library in full. Are there any other changes I should implement to update the library to add this feature?

### DIFF
--- a/tests/response/Vitodens050W_2.json
+++ b/tests/response/Vitodens050W_2.json
@@ -1,0 +1,1112 @@
+{
+    "data": [
+        {
+            "apiVersion": 1,
+            "commands": {
+                "setMode": {
+                    "isExecutable": true,
+                    "name": "setMode",
+                    "params": {
+                        "mode": {
+                            "constraints": {
+                                "enum": [
+                                    "comfort",
+                                    "eco",
+                                    "off"
+                                ]
+                            },
+                            "required": true,
+                            "type": "string"
+                        }
+                    },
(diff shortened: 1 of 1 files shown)


CFenner:
Hey @joearkay thanks for providing the data! It would be nice it you update the existing file, that way we can see what changed. Note that the file [needs to be sorted](https://github.com/openviess/PyViCare/blob/master/.github/workflows/format.yml) to do so.

CFenner:
I already started exposing the *new* function here https://github.com/openviess/PyViCare/pull/427 but as I do not have a DHW module, the motivation is not that high.

joearkay:
Hi @CFenner , Ok thanks. I'll close the PR and add a new one overwriting the same file. Aside, I was experimenting with the library yesterday and noticed that the` setDomesticHotWaterOperatingMode()` does actually interact with the `heating.dhw.operating.modes.active` API. It's just that there is only a 'setter' and no matching 'getter'. So with the exception of the 'getter' function, I guess my 'feature request' is null & void.

CFenner:
Reusing this would also be fine. Just rename the file.

PR #486, opened 2025-01-20
chore(testdata): Update Vitodens 050W data

joearkay:
Overwritten & sorted file. Unable to re-use previous PR as got my branches in a muddle. Sorry.


CFenner (commented):
Can you please look at the indentation, that seems to be a different one.

CFenner:
@joearkay I did the sorting myself, as you can see there are not that many changes.You seem to have no circulation pump active, but instead some DHW modes.

We need to adapt the tests to you values though.

joearkay:
@CFenner thanks, you got to it faster than I did. Ok great, I will have a look

PR #557, opened 2025-05-30
feat: add floor heating + room control

Roguyt:


lackas:
Hi @Roguyt, I came across this PR while working on related PyViCare changes. Nice work — the room control API surface is well covered (sensors, heating/cooling programs, schedules, quick modes) and having real device test data for 6 rooms is very helpful.

Are you still working on this? I'd be happy to help get it across the finish line if you want.

A few things I noticed while reading through the diff:

- `DeviceWithComponent` duplicates `self.service` from the parent device rather than inheriting from `Device`, so Room/Actor don't get any of Device's methods. Might be worth rethinking that relationship.
- The 6 program blocks in Room (normalHeating, reducedHeating, comfortHeating + cooling equivalents) are near-identical — could be reduced with a helper method, similar to how `HeatingDevice` handles circuits.
- `Actor` class is empty beyond a property alias — is there more planned there?
- `FloorHeatingCircuitChannel` is added as a new file but isn't wired into auto-detect. Intentional?
- Small regex bug: `E3_FloorHeatingCircuitDistributorBox` appears twice in the FloorHeating pattern.

None of these are blockers, just things that stood out. Let me know if you'd like a hand with the conflict resolution or any of the above.

lackas:
@CFenner I just came across the discussion in the ViCare Discord about rooms and your reaction to it. I hadn't seen that before.

I understand the HA team doesn't want rooms mapped as devices, but looking at the test data here, the sensor readings (temperature, humidity, CO2) are only available on the RoomControl virtual device under `rooms.N.sensors.*` — the physical Zigbee sensors only expose battery level to HA. So there's no way to get temperature data without going through RoomControl.

A few approaches come to mind:

1. **Circuits pattern** — RoomControl as one HA device, rooms as sub-components (like heating circuits). Entities per room, users assign to HA areas.
2. **Enrich physical devices** — cross-reference RoomControl actors to find which Zigbee sensor is in which room, attach the room's temperature data to the physical sensor device.
3. Something else?

What's your thinking on this? I'd be happy to help move it forward once we know the direction.

CFenner:
Hey, I had in mind to go with 2., exposing the temp (get/set) on the device via the room control.

Roguyt:
Hey @lackas,

Sorry I didn't get the notification.

It's a bit more complicated then that because Viessmann used to not expose many of the stuff I'm using through the free tier API. This changed when they were slapped by EU to release our data basically. Which also meant that prior to that changes you had to be a German customer, willing to pay the stupid 20e fees per month.

TL;DR from my memory of when  I was working on it (and honestly it's more or less that I'm not great with Python combined with a lack of motivation, since I'm a nodejs/ts guy and I'm not changing the temperature that often).

As of 2025.8 (I haven't moved to latest yet), "sensors" (humidity, battery level and temp) are reported separately, and working perfectly fine after @CFenner used my dumps to integrate it.
What's not working, is that as long as you don't implement the room control, you can't change the temperature requested for the given room.
The floor heating part, should report the valve position, but IIRC, it only reported its name so...

Main priority should be to port the room stuff, but at the same time HA doesn't really have such a concept so I'm not too sure how to port it later on to the HA integration.
But at the same time, when i digged through the API, I'd say 80% of what vicare can do, you can't do it through the API. Like the room stuff is really basic and the floor heating thing is even worst.

image

Feels free to dm me on Discord: @roguyt for dumps or anything you need to test on my installation.

Roguyt:
> Hey, I had in mind to go with 2., exposing the temp (get/set) on the device via the room control.

The issue with that is that "technically" you can have 2 climate sensor in the same room, so you would be duplicating the get/set command :/

PR #560, opened 2025-06-08
feat: Add target temperature for heating circuits

bartcortooms:


CFenner (changes requested):
@bartcortooms I guess we can merge this once it's cleaned up.

Issue #452, opened 2024-11-11
getDomesticHotWaterOperatingMode ?


redlegoman:
There seems to be a "setDomesticHotWaterOperatingMode" (to set it to "eco" or "comfort" in my case) but no function to show the current setting. Or is it called something less obvious than "getDomesticHotWaterOperatingMode"?

CFenner:
there was a change to the way DHW is handled in the API back then in November 2022: https://documentation.viessmann.com/static/changelog

This is not yet implemented in the library. I recently started to add them here: https://github.com/openviess/PyViCare/pull/427

lackas:
This is available as `getDomesticHotWaterActiveOperatingMode()`, which reads `heating.dhw.operating.modes.active` and returns the active mode (e.g. `comfort`/`eco` on your device). `getDomesticHotWaterOperatingModes()` gives the selectable modes and `setDomesticHotWaterOperatingMode()` is the setter you already found. The getter name is just less obvious than `getDomesticHotWaterOperatingMode`. Suggest closing.

PR #575, opened 2025-07-11
feat(dhw): expose operation mode

redlegoman:
added getDomesticHotWaterOperatingMode which compliments setDomesticHotWaterOperatingMode. Returns "eco" or "comfort".


CFenner (commented):
Please remove the value from the [ignore list](https://github.com/openviess/PyViCare/blob/eb922bb3dc0174c06c019276e09fdc1e635f3250/tests/test_TestForMissingProperties.py#L49) in `TestForMissingProperties` and add a test case (e.g. https://github.com/openviess/PyViCare/blob/eb922bb3dc0174c06c019276e09fdc1e635f3250/tests/test_VitovalorPT2.py#L65-L66).

PyViCare/PyViCareHeatingDevice.py
```diff
@@ -162,6 +162,10 @@ def getDomesticHotWaterMinTemperature(self):
     def getDomesticHotWaterChargingActive(self):
         return self.getProperty("heating.dhw.charging")["properties"]["active"]["value"]
 
+    @handleNotSupported
+    def getDomesticHotWaterOperatingMode(self):
+        return self.service.getProperty("heating.dhw.operating.modes.active")["properties"]["value"]["value"]
+
```
CFenner, line 168: I think we should update the `getDomesticHotWaterActiveMode` function.

lackas:
master already has `getDomesticHotWaterActiveOperatingMode()`, which reads the same `heating.dhw.operating.modes.active` value, so this getter duplicates existing functionality. If the shorter name is preferred it could be added as an alias to that method; otherwise this can be closed. See #452.

PR #608, opened 2025-09-24
feat(ess): add access methode for ess.transfer.charge.cumulated

lgraf:
Add access method to charge statistics of the vx3 battery. The feature `ess.transfer.charge.cumulated` was finally added with the eu data act changes.

SEE: https://api.viessmann-climatesolutions.com/documentation/data-points

### DIFF
--- a/tests/response/VitochargeVX3.json
+++ b/tests/response/VitochargeVX3.json
@@ -71,6 +71,44 @@
       "uri": "https://api.viessmann-climatesolutions.com/iot/v1/equipment/installations/#######/gateways/################/devices/0/features/ess.stateOfCharge"
     },
     {
+      "feature": "ess.transfer.charge.cumulated",
+      "gatewayId": "################",
+      "deviceId": "0",
+      "timestamp": "2025-09-24T16:25:08.243Z",
+      "isEnabled": true,
+      "isReady": true,
+      "apiVersion": 1,
+      "uri": "https://api.viessmann-climatesolutions.com/iot/v1/features/installations/#######/gateways/################/devices/0/features/ess.transfer.charge.cumulated",
+      "properties": {
+        "currentDay": {
+          "type": "number",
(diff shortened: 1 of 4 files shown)


tests/response/VitochargeVX3.json
CFenner, line 1: Please update the whole file if you don't mind. That way we can see if something else changed.

And please sort the test data as mentioned here: https://github.com/openviess/PyViCare?tab=readme-ov-file#more-different-devices-for-test-cases-needed

lgraf: @CFenner i added a complete snapshot of my device. There are indeed several new values that are interesting, like string voltage, string current, ambient temperature. :)

To bring this feature forward, i think we can introduce the other features in seperate PRs.

Just to let you know i already running a PyViCare version of this branch with an updated vicare integration in my production home assist installation (SEE: https://github.com/lgraf/core/commit/1c0b233a856e88479587e28801c7f1ea4c7af0c2). If a new PyViCare version is released, i can create the PR for the vicare integration .

CFenner: Thanks, now unfortunately the updated data makes the tests fail, because the values changed. Can you adjust that 🙈 ?

lgraf: I see, I will take care tomorrow, didn't run tests locally before commit 🙈

tests/test_TestForMissingProperties.py
```diff
@@ -18,6 +18,13 @@ def test_missingProperties(self):
         # are added to the response files
 
         ignore = [
+            # general - not yet used
```
lgraf, line 21: These properties seems to be new and general (a response for my heatpump also contains these).

CFenner:
https://github.com/home-assistant/core/pull/153631

PR #626, opened 2025-10-19
feat: fetch data via gateway

CFenner:
This is an attempt to optimize the data update mechanism to get more frequent updates when multiple devices present on an account (e.g. zigbee).


CFenner:
superseeded by #777

PR #634, opened 2025-10-24
feat: add support for using client id extracted from android app

reey:
By setting this scope I'm able to use the clientId I've been able to extract from the android app.

This should hopefully allow to use this integration without ratelimits..
I can provide the clientId I've been able to extract on request.

### DIFF
--- a/PyViCare/PyViCareBrowserOAuthManager.py
+++ b/PyViCare/PyViCareBrowserOAuthManager.py
@@ -52,7 +52,7 @@ class ViCareBrowserOAuthManager(AbstractViCareOAuthManager):
         oauth_session = OAuth2Session(
             self.client_id, redirect_uri=redirect_uri, scope=VIESSMANN_SCOPE, code_challenge_method='S256')
         code_verifier = generate_token(48)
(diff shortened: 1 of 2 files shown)


lackas:
We've run both scopes side by side on the same account: the Internal scope does expose a lot more (roughly 2.4x the features here, including room actor mapping and per-module battery data), so the motivation is real. It is not a strict superset, though: a few dozen IoT features are absent under Internal (mostly per-room config flags, plus some VitoCharge PCC power values).

Two concerns with this change as written:
1. It hardcodes `Internal` for everyone. That scope is only granted to the ViCare app's client id; a normal developer-portal client id (what most users and HA Core use) is not authorized for it, so forcing the scope breaks the standard flow. It should be opt-in, selected together with the app client id, not the default.
2. That app client id is undocumented and extracted from the app, so Viessmann can rotate or revoke it anytime. Downstream, HA Core's "official APIs only" policy makes this unlikely to land in the mainline integration; a custom component is the realistic home for it.

Keeping the current default and adding an opt-in path (Internal scope + app client id) would let people who want it use it without regressing everyone else.

PR #698, opened 2026-02-03
feat: expose device_type on PyViCareDeviceConfig

### LINKED ISSUE #565: Order of devices changed
Since a couple of days my values where not reported any more. After some debugging I found out that either a new device is reported by Viessmann API or the order has changed

```
"devices": [
                        {
                            "gatewaySerial": "....",
                            "id": "gateway",
                            "boilerSerial": null,
                            "boilerSerialEditor": null,
                            "bmuSerial": null,
                            "bmuSerialEditor": null,
                            "createdAt": "2025-06-10T11:30:26.939Z",
                            "editedAt": "2025-06-10T11:30:26.939Z",
                            "modelId": "Heatbox1",
                            "status": "Online",
                            "deviceType": "vitoconnect",
                            "roles": [
                                "type:gateway;VitoconnectOpto1",
                                "type:legacy"
                            ],
                            "isBoilerSerialEditable": false,
                            "brand": null,
                            "translationKey": "Heatbox1"
                        },
                        {
                            "gatewaySerial": "...",
                            "id": "0",
                            "boilerSerial": "...",
                            "boilerSerialEditor": "DeviceCommunication",
                            "bmuSerial": "...",
                            "bmuSerialEditor": "DeviceCommunication",
                            "createdAt": "2025-06-10T11:30:26.939Z",
                            "editedAt": "2025-06-10T11:31:28.401Z",
                            "modelId": "CU401B_S",
                            "status": "Online",
                            "deviceType": "heating",
                            "roles": [
                                "capability:consumptionReport;electric",
                                "capability:monetization;AdvancedReport",
                                "capability:productionReport;electric",
                                "capability:productionReport;thermal",
                                "capability:service;AdvancedReport",
                                "type:brand;Viessmann",
                                "type:cooling;integrated",
                                "type:dhw;integrated",
                                "type:heating;integrated",
                                "type:heatpump",
                                "type:legacy",
                                "type:product;CU401B"
                            ],
                            "isBoilerSerialEditable": false,
                            "brand": "Viessmann",
                            "translationKey": "CU401B_S_22"
                        }
```

In the example it is proposed to always take the device[0] which is now the opto-connect-gate way and not the heating anymore (https://github.com/openviess/PyViCare/blob/master/README.md?plain=1#L74)

Either simply hardcode the correct index or use something like this to ignore the "Heatbox1"-Gateways:

```
for device in vicare.devices:
    if (device.device_model != "Heatbox1" ):
        break
```

lackas:
## Summary
- Pass `deviceType` from API response through to `PyViCareDeviceConfig`
- Add `getDeviceType()` method and `device_type` property
- Add `isGateway()` convenience method

## Motivation
Fixes #565 - The Viessmann API may return devices in unpredictable order. Users with multiple devices (e.g., gateway + heating) using `devices[0]` may get unexpected results when the gateway appears first.

With this change, users can filter devices by type:
```python
heating_devices = [d for d in vicare.devices if not d.isGateway()]
# or
heating_devices = [d for d in vicare.devices if d.getDeviceType() == "heating"]
```

## Test plan
- Added 5 unit tests for `getDeviceType()` and `isGateway()`
- All 516 tests pass

### DIFF
--- a/PyViCare/PyViCareDeviceConfig.py
+++ b/PyViCare/PyViCareDeviceConfig.py
@@ -91,7 +91,7 @@ class PyViCareDeviceConfig:
         return self.device_type
 
     def isGateway(self):
-        return self.device_type == "vitoconnect"
+        return self.device_type in ["vitoconnect", "tcu"]
 
     # see: https://vitodata300.viessmann.com/vd300/ApplicationHelp/VD300/1031_de_DE/Ger%C3%A4teliste.html
     def asAutoDetectDevice(self):
--- a/PyViCare/PyViCareDeviceConfig.py
+++ b/PyViCare/PyViCareDeviceConfig.py
@@ -22,11 +22,12 @@ logger.addHandler(logging.NullHandler())
 
 
 class PyViCareDeviceConfig:
-    def __init__(self, service, device_id, device_model, status):
+    def __init__(self, service, device_id, device_model, status, device_type=None):
         self.service = service
         self.device_id = device_id
         self.device_model = device_model
         self.status = status
+        self.device_type = device_type
 
     def asGeneric(self):
         return HeatingDevice(self.service)
@@ -84,6 +85,12 @@ class PyViCareDeviceConfig:
 
     def isOnline(self):
         return self.status == "Online"
+
+    def getDeviceType(self):
+        return self.device_type
+
+    def isGateway(self):
+        return self.device_type == "vitoconnect"
 
     # see: https://vitodata300.viessmann.com/vd300/ApplicationHelp/VD300/1031_de_DE/Ger%C3%A4teliste.html
     def asAutoDetectDevice(self):
--- a/PyViCare/PyViCareDeviceConfig.py
+++ b/PyViCare/PyViCareDeviceConfig.py
@@ -22,6 +22,7 @@ logger.addHandler(logging.NullHandler())
 
 
 class PyViCareDeviceConfig:
+    # pylint: disable=too-many-arguments,too-many-positional-arguments
     def __init__(self, service, device_id, device_model, status, device_type=None):
         self.service = service
         self.device_id = device_id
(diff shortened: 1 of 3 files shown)


lackas:
Fixed the pylint failure by adding an inline disable comment.

If you'd prefer a different approach, I'm happy to refactor to use a `@dataclass` or pass parameters via a config object instead. Let me know!

PyViCare/PyViCareDeviceConfig.py
```diff
@@ -85,6 +87,12 @@ def getModel(self):
     def isOnline(self):
         return self.status == "Online"
 
+    def getDeviceType(self):
+        return self.device_type
+
+    def isGateway(self):
+        return self.device_type == "vitoconnect"
```
CFenner: Are you sure this is enough? Are there no other gateway types?

lackas (commented):
Good catch! The TCU (Telematic Control Unit) series are also gateways — the API returns `deviceType: "tcu"` for those. Updated `isGateway()` to check for both:

```python
def isGateway(self):
    return self.device_type in ["vitoconnect", "tcu"]
```

This aligns with the existing filter in `PyViCare.py:55` which already excludes both `"vitoconnect"` and `"tcu"` device types.

CFenner:
Do we need to align with https://github.com/openviess/PyViCare/blob/10328d412475024ef74415c0c7c40e669581794a/PyViCare/PyViCareService.py#L50-L51

lackas (commented):
Good point! Delegated to the existing role-based `_isGateway()` check on the service, so there's a single source of truth covering all gateway types (VitoconnectOpto1/2, TCU100/200/300).

Note: this requires a `pylint: disable=protected-access` since `_isGateway` is technically private. We could also refactor `_isGateway` to be public on the service, but I wanted to keep this PR's scope minimal. Happy to do that in a follow-up if you prefer.

CFenner (approved):

PR #704, opened 2026-02-09
fix: handle DEVICE_COMMUNICATION_ERROR and return stale cache on transient errors

### LINKED ISSUE #361: DEVICE_COMMUNICATION_ERROR
From time to time I see `PyViCareInvalidDataError` in HA. It would be nice if this would be detected as gateway being offline and setting the state respectively instead of failing.

```
Invalid data from Vicare server: {'viErrorId': '|00-bc04a79ce06442efa7a77de4e7287029-f22ac2b8491e4b09-01.d35d83cd_', 'statusCode': 400, 'errorType': 'DEVICE_COMMUNICATION_ERROR', 'message': '', 'extendedPayload': {'httpStatusCode': 'NotFound', 'code': '404', 'reason': 'GATEWAY_OFFLINE'}}
```

lackas:
## Summary

When the Viessmann API returns transient errors (device offline, gateway offline, server errors), PyViCare either raises a misleading `PyViCareInvalidDataError` or crashes callers that don't catch `PyViCareInternalServerError`.

This PR:
1. Adds `PyViCareDeviceCommunicationError` for API responses with `errorType == "DEVICE_COMMUNICATION_ERROR"`
2. Detects these in `get()` alongside existing server error handling
3. Returns stale cache in `PyViCareCachedService` when a transient error occurs and cached data is available, only raising if no cache exists

The rate limiting behavior (one API call per cache duration) is preserved — only the caller-facing behavior changes from "raise exception" to "return stale data" when cached data is available.

Fixes #361

## Changes

- **`PyViCareUtils.py`**: New `PyViCareDeviceCommunicationError` exception with `reason` attribute
- **`PyViCareAbstractOAuthManager.py`**: Detect `DEVICE_COMMUNICATION_ERROR` in `get()`
- **`PyViCareCachedService.py`**: Catch transient errors, return stale cache when available
- **Tests**: 10 new tests covering exception construction, OAuth manager detection, and cached service stale cache behavior

### DIFF
--- a/PyViCare/PyViCareAbstractOAuthManager.py
+++ b/PyViCare/PyViCareAbstractOAuthManager.py
@@ -62,7 +62,7 @@ class AbstractViCareOAuthManager:
             raise PyViCareRateLimitError(response)
 
     def __handle_device_communication_error(self, response):
-        if response.get("errorType") == "DEVICE_COMMUNICATION_ERROR":
+        if ("errorType" in response and response["errorType"] == "DEVICE_COMMUNICATION_ERROR"):
             raise PyViCareDeviceCommunicationError(response)
 
     def __handle_server_error(self, response):
--- a/PyViCare/PyViCareAbstractOAuthManager.py
+++ b/PyViCare/PyViCareAbstractOAuthManager.py
@@ -7,6 +7,7 @@ from authlib.integrations.requests_client import OAuth2Session
 
 from PyViCare import Feature
 from PyViCare.PyViCareUtils import (PyViCareCommandError,
+                                    PyViCareDeviceCommunicationError,
                                     PyViCareInternalServerError,
                                     PyViCareRateLimitError)
 
@@ -39,6 +40,7 @@ class AbstractViCareOAuthManager:
             logger.debug("Response to get request: %s", response)
             self.__handle_expired_token(response)
             self.__handle_rate_limit(response)
+            self.__handle_device_communication_error(response)
             self.__handle_server_error(response)
             return response
         except TokenExpiredError:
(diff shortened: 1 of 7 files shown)


lackas:
Companion HA Core PR (draft, waiting for this to merge + version bump): home-assistant/core#162618

PyViCare/PyViCareAbstractOAuthManager.py
```diff
@@ -59,6 +61,10 @@ def __handle_rate_limit(self, response):
         if ("statusCode" in response and response["statusCode"] == 429):
             raise PyViCareRateLimitError(response)
 
+    def __handle_device_communication_error(self, response):
+        if response.get("errorType") == "DEVICE_COMMUNICATION_ERROR":
```
CFenner: Lets check the key here as well:

```suggestion
        if ("errorType" in response and response.get("errorType") == "DEVICE_COMMUNICATION_ERROR"):
```

PyViCare/PyViCareCachedService.py
```diff
@@ -33,13 +36,20 @@ def setProperty(self, property_name, action, data):
...
                 # we simply return the old cache in this case
                 self.__cacheTime = ViCareTimer().now()
 
-                data = self.fetch_all_features()
+                try:
+                    data = self.fetch_all_features()
+                except (PyViCareDeviceCommunicationError, PyViCareInternalServerError) as e:
+                    if self.__cache is not None:
+                        logger.warning("API error, returning stale cache: %s", e)
+                        return self.__cache
```
CFenner, line 50: This way a user will no longer get the exception when a device is offline. So, how will he know?

PyViCare/PyViCareAbstractOAuthManager.py
lackas: Good point, updated to match the existing pattern.

PyViCare/PyViCareCachedService.py
lackas: The `logger.warning("API error, returning stale cache: %s", e)` logs it, so it will appear in the logs. And when there is no stale cache yet (e.g. first call after startup), the exception is re-raised so the caller still sees it.

The main use case is Home Assistant, where returning stale data is much better than crashing all entities to unavailable on a transient gateway hiccup. HA will also handle the new `PyViCareDeviceCommunicationError` explicitly in the integration code (see companion PR home-assistant/core#162618).

CFenner (approved):
👍

PR #689, opened 2026-01-22
feat(heat pump): add COP, compressor sensors, refrigerant circuit and heating rod stats

lackas:
## Summary

Add comprehensive heat pump monitoring methods, primarily for Vitocal 300-G but applicable to other models exposing these API features.

**New features:**
- COP (Coefficient of Performance): heating, DHW, total, cooling, green
- Compressor: power, modulation, load class fallback for `statistics.load` path
- Refrigerant circuit: hot gas/suction gas pressure & temperature, liquid gas temperature
- Heating rod: runtime by level, power consumption summary (DHW & heating)
- Configuration: buffer temp max, damping factor, heater approvals
- Primary circuit pump rotation
- Circuit target temperature
- New `CoolingCircuit` class with type and reverse active status

Closes #677

## Review guide

**Total: ~290 lines of library code, ~280 lines of tests**

| Lines | Type | What to review |
|-------|------|----------------|
| ~35 | Logic | Load class fallback in `Compressor` (lines 547-585) - tries `statistics` then `statistics.load` |
| ~15 | New class | `CoolingCircuit` class (lines 529-543) |
| ~240 | Boilerplate | Standard getters following existing pattern: `@handleNotSupported` decorator + single-line property access |

The boilerplate getters all follow the exact same pattern as existing methods like `getSeasonalPerformanceFactorHeating()`. You can skim these quickly - they're just exposing new API properties.

## Notes

- Rebased on master after #697 (test data) was merged
- Squashed to single commit for cleaner review
- All 61 Vitocal300G tests now in one file (consolidated from separate CU401B file)

## Test plan

- [x] 61 tests in `test_Vitocal300G.py` covering all new methods
- [x] 16 tests in `test_Vitocal250A.py` for heating rod power consumption summary
- [x] Updated ignore list in `test_TestForMissingProperties.py`
- [x] All 563 tests pass
- [x] Ruff passes

Issue #363, opened 2024-03-15
Natural cooling


jchapuis:
A proposal to surface the endpoint 

https://api.viessmann.com/iot/v2/features/installations/{installation_id}/gateways/{gateway_id}/devices/0/features/heating.circuits.1.operating.modes.cooling 

Afaik this returns a boolean flag that allows to determine whether natural cooling is currently running, which can be helpful to automate opening up valves

CFenner:
There is already the `getActiveProgram` which you can already use to compare yourself.

https://github.com/somm15/PyViCare/blob/6c5b023ca6c8bb2d38141dd1746dc1705ec84ce8/PyViCare/PyViCareHeatingDevice.py#L591-L593

One could also add a generic function like this to handle it for all programs: 

https://github.com/somm15/PyViCare/blob/6c5b023ca6c8bb2d38141dd1746dc1705ec84ce8/PyViCare/PyViCareHeatingDevice.py#L459-L464

jchapuis:
thanks! actually, these endpoints do not seem to deliver enough details. What I'm looking for are not settings but whether the circulation pump is currently pumping heat down into the ground, so that I can open up all the floor heating valves.

```
"uri": "https://api.viessmann.com/iot/v2/features/installations/{}/gateways/{}/devices/0/features/heating.circuits.1.operating.programs.active",
        "properties": {
            "value": {
                "type": "string",
                "value": "reduced"
            }
        },
```

```
 "uri": "https://api.viessmann.com/iot/v2/features/installations/{}/gateways/{}/devices/0/features/heating.circuits.1.operating.modes.active",
        "properties": {
            "value": {
                "type": "string",
                "value": "dhwAndHeatingCooling"
            }
        },
```

CFenner:
Sorry, I mixed up program and mode. But there is also a function `getActiveMode`. 

Maybe your problem can be solved with comparing `temperature.supply` and `temperature.return` sensor values?

jchapuis:
> Sorry, I mixed up program and mode. But there is also a function `getActiveMode`.
> 
> Maybe your problem can be solved with comparing `temperature.supply` and `temperature.return` sensor values?

good hint! i could also compound this with the circulation pump running, as natural cooling only uses the pump. However, since the API seems to have this flag to indicate cooling operation currently running, would have been nice to surface this. 
[...]

PR #706, opened 2026-02-10
feat: add CoolingCircuit support for heat pumps

lackas:
Add detection and accessors for cooling circuits on heat pumps with cooling capability.

Split out from #689 per review feedback to keep PRs focused.

**Changes:**
- `CoolingCircuit` class with `getType()` and `getReverseActive()`
- Auto-detection of available cooling circuits (0-3) via `getAvailableCoolingCircuits()`
- `coolingCircuits` property on `HeatPump`
- Tests for Vitocal 300G CU401B

**Tested devices** (feature present in test data):
- Vitocal 300G CU401B
- Vitocal 333G with Vitovent 300F
- Vitocal 200S AWB-M-E-AC-201.D10

**Deprecation check:** Verified all cooling circuit features (`heating.coolingCircuits.*.type`, `heating.coolingCircuits.*.reverse`) against the API's `deprecated` field in recent device dumps — none are flagged as deprecated.

Addresses part of #363 (natural cooling).

### DIFF
--- a/PyViCare/PyViCareHeatPump.py
+++ b/PyViCare/PyViCareHeatPump.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
+from contextlib import suppress
 from typing import Any, List
 from deprecated import deprecated
 
 from PyViCare.PyViCareHeatingDevice import HeatingDevice, HeatingDeviceWithComponent
-from PyViCare.PyViCareUtils import handleAPICommandErrors, handleNotSupported
+from PyViCare.PyViCareUtils import (PyViCareNotSupportedFeatureError,
+                                    handleAPICommandErrors, handleNotSupported)
 from PyViCare.PyViCareVentilationDevice import VentilationDevice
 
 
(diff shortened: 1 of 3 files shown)


CFenner:
Out of curiosity, this is something different than pumping cold water through a usual heating circuit?

CFenner (approved):
👍

lackas:
Good question! There are two types of cooling in heat pumps:

**Passive cooling** (Viessmann calls it "natural cooling") — only available with ground-source heat pumps. The ground water (around 10-15°C) is pumped through the heat exchanger to cool the buffer, which then cools the heating circuits. The compressor stays off, so it only uses energy for the pumps.

**Active cooling** — the compressor runs in reverse. Any heat pump that supports reverse operation can do this, but it uses significantly more energy since the compressor is running.

In both cases you have to be careful not to cool floors below the dew point (condensation). The `coolingCircuit` features in the API track the type and whether reverse (active) cooling is currently active.

PR #696, opened 2026-01-25
Update load class methods to use new statistics path

firewiremb:
Along with the API v2 the structure of the 'load class' information has changed. See also https://github.com/openviess/PyViCare/issues/650

PR #695, opened 2026-01-25
feat(ventilation): expose exhaust and extract temperatures and humidities plus heat recovery efficiency

norschel:
This pull request enhances the ventilation device support by adding new sensor readings, updating available programs, and expanding test coverage to reflect these changes. The main focus is on improving the data available for ventilation devices, particularly for the Vitoair FS300E model.

**Ventilation Device Feature Enhancements:**

* Added new sensor reading methods to `PyViCareVentilationDevice.py` for exhaust and extract temperatures and humidities, as well as heat recovery efficiency. These methods are now available for use and tested for support. [[1]](diffhunk://#diff-22b074a82f74d5b867117bb2dd826a224e95672ad220ede142297102babed885R233-R252) [[2]](diffhunk://#diff-22b074a82f74d5b867117bb2dd826a224e95672ad220ede142297102babed885R301-R304)
* Updated the list of available ventilation programs to include the `active` program in `getVentilationPrograms`.

**Test Coverage and Data Updates:**

* Expanded the test suite in `test_VitoairFs300E.py` to cover the new sensor methods and updated expected values for ventilation programs and volume flows to match the new device data. [[1]](diffhunk://#diff-7e3ae88ba5b67ff6cd9b2c398da128ade5b4a44db3b6b153831532f93e6ff8d3L25-R32) [[2]](diffhunk://#diff-7e3ae88ba5b67ff6cd9b2c398da128ade5b4a44db3b6b153831532f93e6ff8d3R59-R84)
* Extended the list of properties checked in `test_TestForMissingProperties.py` to include additional ventilation features observed in the Vitoair FS300E device, improving coverage for new and existing features.This pull request extends the ventilation device support by adding new sensor readings and their corresponding tests, as well as updating the test response data to include these new features. The main changes are the addition of methods for retrieving exhaust and extract temperatures and humidities, as well as heat recovery efficiency, ensuring the library can access more detailed ventilation data.

The additional sensor data is needed for extending the HA Vicare component. 

**Ventilation sensor and efficiency support:**

[...]

PR #667, opened 2025-11-20
feat: expose power consumption for heating, dhw and total

tlinnet:
https://github.com/openviess/PyViCare/issues/664

PR #645, opened 2025-11-06
chore(testdata): add Vitocal 300G sample data

CFenner:
New Viessmann EUDataAct-Data Points are missing
#153184

PR #707, opened 2026-02-10
chore: add deprecation database to prevent use of deprecated API paths

lackas:
## Summary

This adds a deprecation database and tooling to prevent accidentally adding code that uses deprecated Viessmann API paths — saving review effort by catching these issues before a PR is submitted.

### Background

During review of #689, @CFenner pointed out that some features might be deprecated. Looking into it, I found that I had accidentally added a getter for `heating.cop.green`, which is deprecated (replaced by `heating.cop.photovoltaic`, removal 2025-12-31). The existing `test_deprecatedProperties` test didn't warn me because the test response data for that device doesn't include the `deprecated` marker — it was only visible in a fresh API dump. I wanted to improve this so deprecated features are caught automatically before submitting a PR.

### What this adds

**`tests/deprecated_features.json`** — A persistent database of 106 known deprecated features, merged from all test response files and fresh device dumps. Each entry records:
- `removalDate` and replacement info from the API
- `firstSeenIn` / `firstSeenOn` — provenance tracking (which file first reported it, when)
- `sources` — all files that confirm the deprecation

**`check_deprecations.py`** — A developer tool that:
- Reports deprecated features used in code (grouped, with removal dates and replacements)
- Keeps the database up to date from test data (`--update`)
- Ingests fresh device dumps (`--update dump.json`) and highlights newly discovered deprecations
- Collapses repeated entries (e.g. `rooms.*.sensors.window.openState (24x)`)

**Enhanced `test_deprecatedProperties`** — Now also checks the database, so deprecations known from any source (not just what's marked in test data) are caught in CI.

### How fresh dumps help

A feature deprecated in one API response is deprecated everywhere — but not all test data files include the `deprecated` marker (older dumps, or devices that don't expose it). Fresh dumps, even for devices that already have test data, can reveal deprecations missing from existing data. You don't need to commit the full dump — just run:

```
python3 check_deprecations.py --update path/to/dump.json
```

This extracts only the deprecation info into the database. I checked all dumps attached to open PRs (#645, #667, #695, #696) and they don't contain new deprecation info beyond what's already in the database.

### Example output

```
=== WARNING: Deprecated features used in code ===

  ventilation.operating.programs.comfort
    Removal: 2024-09-15 (PAST DUE)
    Used in code: PyViCareVentilationDevice.py
  ...

=== Past removal date (not used in code): 101 features ===

  heating.cop.green
    Removal: 2025-12-31 (PAST DUE)
    Replaced by: replaced by heating.cop.photovoltaic
    From dump: device_0_features.json

  rooms.*.configuration.window.openState (24x)
    Removal: 2024-09-15 (PAST DUE)
    From dump: device_RoomControl-1_features.json
  ...

=== Summary ===
106 deprecated features in database
5 used in code (ACTION NEEDED)
101 past removal date
0 upcoming
```

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -33,7 +33,7 @@ jobs:
     name: validate deprecation database
     runs-on: ubuntu-latest
     steps:
-      - name: Check out code from GitHub
+      - name: ⤵️ Check out code from GitHub
         uses: actions/checkout@v6.0.2
       - name: Set up Python
         uses: actions/setup-python@v6.2.0
@@ -43,3 +43,4 @@ jobs:
         run: python check_deprecations.py --update
       - name: Verify no changes
         run: git diff --name-only --exit-code
+        if: always()
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -28,3 +28,18 @@ jobs:
             -exec sh -c 'mv $1 $1.tmp; jq ".data|=sort_by(.feature)" --sort-keys $1.tmp > $1; rm $1.tmp' shell {} ";"
       - name: 🔍 Verify
         run: git diff --name-only --exit-code
+
+  deprecations:
+    name: validate deprecation database
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code from GitHub
+        uses: actions/checkout@v6.0.2
+      - name: Set up Python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: ${{ env.DEFAULT_PYTHON }}
+      - name: Update deprecation database
+        run: python check_deprecations.py --update
+      - name: Verify no changes
+        run: git diff --name-only --exit-code
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -42,5 +42,5 @@ jobs:
       - name: Update deprecation database
         run: python check_deprecations.py --update
       - name: Verify no changes
-        run: git diff --name-only --exit-code
+        run: git diff --exit-code
         if: always()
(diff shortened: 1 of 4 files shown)


CFenner:
Is data that is added on a PR validated (--update) automatically? Maybe we can have something similar to this: https://github.com/openviess/PyViCare/blob/master/.github/workflows/format.yml

lackas:
Not yet, but good idea — I'll add a CI workflow similar to `format.yml` that runs `check_deprecations.py --update` and verifies no diff, so new test data automatically keeps the database in sync.

lackas:
The CI job is failing because `check_deprecations.py` correctly detects 5 deprecated features still used in the ventilation code (all past due since 2024-09-15, see #708). Since fixing those requires an interface redesign and someone with ventilation hardware to test, how would you like to proceed? We could use `continue-on-error: true` on the check for now so it's visible but non-blocking until the ventilation migration is done.

CFenner:
Currently git diff fails just because of the "update" changes. I think it would be better to just update, if there are any other changes.

Bildschirmfoto 2026-02-15 um 13 08 16

lackas:
Good point\! Fixed — `save_database()` is now only called when features actually changed, so `--update` with no new deprecations won't touch the file (and `git diff` will pass).

lackas:
The git diff issue is fixed now. The remaining CI failure is from the 5 deprecated ventilation features still used in code (tracked in #708) — `check_deprecations.py` correctly exits with code 1 for those. Should we set this job to `continue-on-error: true` until the ventilation migration is done, or would you prefer a different approach (e.g., an allow-list)?

PR #725, opened 2026-03-18
chore: Expose oauth constants

lackas:
Move `AUTHORIZE_URL`, `TOKEN_URL`, and `VIESSMANN_SCOPE` to `PyViCareAbstractOAuthManager` as the single source of truth. Both `PyViCareOAuthManager` and `PyViCareBrowserOAuthManager` now import from there.

This aligns the scope to `["IoT User", "offline_access"]` for both managers. Previously, the password-grant manager used only `["IoT User"]`, which meant tokens had no refresh capability — they expired after 1h with no way to renew without re-authenticating with username/password.

Adding `offline_access` to the password-grant flow gives it a refresh token too, improving reliability for long-running sessions.

Related: openviess/PyViCare#724, home-assistant/core#165621

### DIFF
--- a/PyViCare/PyViCareAbstractOAuthManager.py
+++ b/PyViCare/PyViCareAbstractOAuthManager.py
@@ -15,6 +15,9 @@ logger = logging.getLogger('ViCare')
 logger.addHandler(logging.NullHandler())
 
 API_BASE_URL = 'https://api.viessmann-climatesolutions.com/iot/v2'
+AUTHORIZE_URL = 'https://iam.viessmann-climatesolutions.com/idp/v3/authorize'
+TOKEN_URL = 'https://iam.viessmann-climatesolutions.com/idp/v3/token'
+VIESSMANN_SCOPE = ["IoT User", "offline_access"]
 
 
 class AbstractViCareOAuthManager:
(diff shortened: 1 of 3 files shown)


PyViCare/PyViCareAbstractOAuthManager.py
```diff
@@ -15,6 +15,9 @@
 logger.addHandler(logging.NullHandler())
 
 API_BASE_URL = 'https://api.viessmann-climatesolutions.com/iot/v2'
+AUTHORIZE_URL = 'https://iam.viessmann-climatesolutions.com/idp/v3/authorize'
+TOKEN_URL = 'https://iam.viessmann-climatesolutions.com/idp/v3/token'
+VIESSMANN_SCOPE = ["IoT User", "offline_access"]
```
CFenner: Lets properly define all available scopes:

```suggestion

SCOPE_IOT = "IoT"
SCOPE_USER = "User"
SCOPE_OFFLINE_ACCESS = "offline_access"
SCOPE_INTERNAL = "internal"
```

CFenner (changes requested):
I would rather replace the usage of `VIESSMANN_SCOPE` with the actual scope lists (`[SCOPE_IOT, SCOPE_OFFLINE_ACCESS]`).

PyViCare/PyViCareOAuthManager.py
```diff
@@ -55,7 +59,7 @@ def __create_new_session(self, username, password, token_file=None):
             oauth sessions object
         """
         oauth_session = OAuth2Session(
-            self.client_id, redirect_uri=REDIRECT_URI, scope=VIESSMANN_SCOPE, code_challenge_method='S256')
+            self.client_id, redirect_uri=REDIRECT_URI, scope=[SCOPE_IOT, SCOPE_USER, SCOPE_OFFLINE_ACCESS], code_challenge_method='S256')
```
CFenner:
```suggestion
            self.client_id, redirect_uri=REDIRECT_URI, scope=[SCOPE_IOT, SCOPE_USER], code_challenge_method='S256')
```

CFenner (approved):

PR #738, opened 2026-04-11
chore: Change logger name to 'PyViCare...'

lackas:
All 7 modules use `logging.getLogger('ViCare')`, but Home Assistant's vicare integration registers `"loggers": ["PyViCare"]` in its `manifest.json`. This mismatch means HA's logger configuration for the integration has no effect on the library's log output.

Renaming to `'PyViCare'` matches the package name (Python convention) and makes HA's logger settings work as expected.

### DIFF
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -9,7 +9,7 @@ from PyViCare.PyViCareOAuthManager import ViCareOAuthManager
 from PyViCare.PyViCareService import ViCareDeviceAccessor, ViCareService
 from PyViCare.PyViCareUtils import PyViCareInvalidDataError
 
-logger = logging.getLogger('ViCare')
+logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
 
--- a/PyViCare/PyViCareAbstractOAuthManager.py
+++ b/PyViCare/PyViCareAbstractOAuthManager.py
@@ -11,7 +11,7 @@ from PyViCare.PyViCareUtils import (PyViCareCommandError,
                                     PyViCareInternalServerError,
                                     PyViCareRateLimitError)
 
-logger = logging.getLogger('ViCare')
+logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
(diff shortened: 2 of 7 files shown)


lackas:
@CFenner Quick question: the library uses `logging.getLogger('ViCare')` but HA's manifest registers `"loggers": ["PyViCare"]`. Was the `ViCare` name intentional (to separate library logs from integration logs), or is the mismatch unintended?

If `ViCare` is intentional, the fix should be on the HA side instead (`manifest.json` → `"loggers": ["ViCare"]`). Happy to adjust either way.

lackas:
Alternatively, I could switch to `logging.getLogger(__name__)` instead of a hardcoded string — that would give per-module loggers (`PyViCare.PyViCareCachedService`, etc.) which makes it easier to trace where a log line came from. Downside is it changes the log output format. Happy to go either way.

CFenner:
I think it is intentional as it matches the integration name. 

Actually I have no idea what the manifest entry is for.

lackas:
The `loggers` field in `manifest.json` tells HA which library loggers to enable when a user turns on debug logging for the integration via the UI. Right now the manifest has `"loggers": ["PyViCare"]`, but the library uses `getLogger('ViCare')` — so enabling debug for the integration gives you integration-level logs but nothing from the library.

I confirmed this on our system: with debug enabled, zero `[ViCare]` lines in the output.

We could fix it on the HA side (change manifest to `"loggers": ["ViCare"]`), but I think using the package name `PyViCare` as the logger namespace is more predictable — it's what you'd expect when looking at the installed package. Happy to go either way though.

CFenner:
Ah, Sorry thought this is the HA repo 🙄 would be nice if you could switch to the more meaningful __name__.

lackas:
Updated to use `__name__` instead of hardcoded `'PyViCare'` — gives per-module granularity (e.g. `PyViCare.PyViCareService`, `PyViCare.PyViCareCachedService`) while still being captured by HA's `"loggers": ["PyViCare"]` manifest entry thanks to Python's hierarchical logging.

CFenner (approved):

PR #740, opened 2026-04-15
fix: auto-detect hybrid devices using API features

lackas:
## Problem

Hybrid devices like the Vitocaldens 222-F (controller `CU401B_S`) are detected as `HeatPump` because `CU401B` matches the heat pump regex pattern. The `Hybrid` class was never in the auto-detect list — it was only reachable via the manual `heating_type: hybrid` config option in Home Assistant, which was removed in home-assistant/core#165649.

This causes all burner entities to disappear, since `HeatPump` doesn't expose `heating.burners.*` features — only `Hybrid` (which inherits from both `GazBoiler` and `HeatPump`) does.

## Solution

After regex/role matching picks `GazBoiler` or `HeatPump`, the device's API features are fetched and checked. If both `heating.burners` and `heating.compressors` features exist, the device is upgraded to `Hybrid`.

This is a fallback — the existing regex matching is unchanged. If the feature fetch fails, the original detection is silently kept.

## Why not full feature-based detection?

Non-heating device types (ventilation, radiator actuators, room sensors, gateways) don't have obvious feature-prefix signatures, so they still need regex/role matching. This approach is the smallest change that catches all hybrid devices — current and future — without maintaining a model name list.

Fixes home-assistant/core#168288

### DIFF
--- a/PyViCare/PyViCareDeviceConfig.py
+++ b/PyViCare/PyViCareDeviceConfig.py
@@ -143,7 +143,7 @@ class PyViCareDeviceConfig:
             has_burners = any(f.startswith("heating.burners") for f in feature_names)
             has_compressors = any(f.startswith("heating.compressors") for f in feature_names)
             return has_burners and has_compressors
-        except Exception:
+        except (KeyError, TypeError, AttributeError, OSError):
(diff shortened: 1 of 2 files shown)


CFenner (approved):
nice catch!

PR #741, opened 2026-04-16
fix: handle non-JSON error responses from API gateway

lackas:
When the Viessmann API has issues (maintenance, overload, DNS problems), several error types propagate as unhandled exceptions with full tracebacks in the Home Assistant logs. All of these should be caught so the cached service can fall back to stale data instead of crashing.

This PR fixes three scenarios:

**1. Non-JSON 5xx responses (e.g. 502 HTML from load balancer)**
Calling `.json()` on an HTML error page raises `ContentTypeError`. Now we check the response content type before parsing.

**2. TIMEOUT errors in `extendedPayload`**
The API sometimes returns `{'extendedPayload': {'code': '500', 'reason': 'TIMEOUT'}}` with empty `errorType` and no `statusCode`. This slipped through all existing error handlers and caused `PyViCareInvalidDataError` ("Missing 'data' property") instead of `PyViCareInternalServerError` — bypassing the stale-cache fallback.

**3. Connection-level errors (DNS timeout, connection refused)**
Network errors raise `OSError` before any HTTP response exists, bypassing all error handlers entirely.

All three are now wrapped as `PyViCareInternalServerError`, which the cached service already handles gracefully (returns stale data + logs a warning).

### DIFF
--- a/PyViCare/PyViCareAbstractOAuthManager.py
+++ b/PyViCare/PyViCareAbstractOAuthManager.py
@@ -44,7 +44,9 @@ class AbstractViCareOAuthManager:
     def get(self, url: str) -> Any:
         try:
             logger.debug(self.__oauth)
-            response = self.__oauth.get(f"{API_BASE_URL}{url}", timeout=31).json()
+            raw_response = self.__oauth.get(f"{API_BASE_URL}{url}", timeout=31)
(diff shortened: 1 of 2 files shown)


CFenner (approved):
👍

PR #742, opened 2026-04-16
chore(testdata): add E3_RoomControl_One_525 test data

lackas:
Adds sanitized RoomControl device data (E3_RoomControl_One_525) with 10 rooms including temperature, humidity, and actor mappings to Zigbee room sensors and floor heating channels.

Data contributed by blob810 from a Vitocal 222-SI installation. Serial numbers are hashed, room names translated to English.

This is a prerequisite for #557 (room control / floor heating support). Submitted separately to keep the feature PR focused on code changes.

PR #743, opened 2026-04-16
feat: enrich Zigbee sensors with room data from RoomControl

lackas:
## Background

Viessmann has removed `device.sensors.temperature` and `device.sensors.humidity` from physical Zigbee room sensors (E3_RoomSensor). Comparing current API responses (Apr 2026) with #557 test data (May 2025), the sensors went from 12 features down to 6 — only battery, zigbee signal, and identification remain.

The temperature and humidity data still exists, but only on the RoomControl virtual device under `rooms.N.sensors.*`. This means `RoomSensor.getTemperature()` and `getHumidity()` silently stopped working for affected installations.

## Design

Rather than mapping Viessmann's room abstraction into a new device hierarchy (as #557 proposed), this PR **enriches the existing physical Zigbee devices** with data from RoomControl — following @CFenner's preferred approach (https://github.com/openviess/PyViCare/pull/557#issuecomment-4257657966).

The cross-referencing works because RoomControl lists "actors" per room — these are the Zigbee device IDs of the physical sensors and floor heating channels in that room. We reverse this mapping: given a Zigbee sensor's device ID, find which room it belongs to, then read that room's sensor data.

### How it works

1. `roomControl` added to device type filter — PyViCare now discovers RoomControl devices
2. After device discovery, `__enrichZigbeeDevices()` builds an actor→room mapping from each RoomControl
3. For each Zigbee device whose ID appears as an actor, the RoomControl reference and room ID are stored on the `PyViCareDeviceConfig`
4. When `asRoomSensor()` creates a RoomSensor, the enrichment is applied
5. `RoomSensor.getTemperature()` transparently reads from RoomControl when enriched, falls back to device-level data for installations without RoomControl (older API or different device models)

### What's exposed

Sensors: temperature, humidity, CO2, condensation risk
Operating state: level, demand, reason
Heating programs: normal/reduced/comfort target temperatures with `setTemperature` commands
Quick mode: manual till next schedule (activate/deactivate/set temperature)
Schedule: read (per room)
Room metadata: name, type

### What's NOT in this PR

- No new HA integration entities — that's a separate HA Core PR once this lands
- No changes to FloorHeatingChannel enrichment — floor heating channels also appear as actors but don't need sensor data. Could be added later for operating state.
- Deprecated features (window.openState, operating.programs.active) are excluded

## Test data

Depends on #742 for the E3_RoomControl_One_525 test data (contributed by blob810, sanitized).

## Breaking changes

None. Existing behavior is preserved — enrichment only activates when a RoomControl device exists. Installations without RoomControl are unaffected.


PyViCare/PyViCareRoomSensor.py
```diff
@@ -1,17 +1,132 @@
...
-from PyViCare.PyViCareUtils import handleNotSupported
+from PyViCare.PyViCareUtils import handleNotSupported, handleAPICommandErrors
 
 
 class RoomSensor(ZigbeeBatteryDevice):
 
+    def __init__(self, service):
+        super().__init__(service)
+        self._room_control = None
+        self._room_id = None
```
CFenner: This can move to class level, no?

PyViCare/PyViCareRoomSensor.py
```diff
@@ -1,17 +1,132 @@
...
         return self.getProperty("device.sensors.temperature")["properties"]["value"]["value"]
 
     @handleNotSupported
     def getHumidity(self):
+        if self._room_control is not None:
+            return self._room_control.getRoomHumidity(self._room_id)
         return self.getProperty("device.sensors.humidity")["properties"]["value"]["value"]
+
+    @handleNotSupported
+    def getCO2(self):
```
CFenner: please add types

lackas:
## HA integration steps

Once this lands and is released, the HA vicare integration needs:

1. **Bump PyViCare version** 
2. **Filter out `roomControl` devices from entity creation** — RoomControl is an implementation detail for enrichment, users shouldn't see it as an HA device. One-line check: skip `device_config.getDeviceType() == "roomControl"`.

That's enough to restore `room_temperature` and `room_humidity` sensors — the existing `GLOBAL_SENSORS` descriptions already call `getTemperature()` / `getHumidity()`, which now work again via enrichment.

For the new features (heating targets, condensation risk, schedules, quick modes), we'd add entity descriptions in a follow-up HA PR.

## Impact and risk

**For installations with RoomControl + Zigbee sensors (blob810's case):** Temperature and humidity are restored on the physical sensor devices. Heating targets, condensation risk, and schedules become available. Confirmed working on a 10-room installation.

**For installations without RoomControl (our test case):** The enrichment finds RoomControl devices with 0 rooms, builds an empty actor map, and does nothing. No extra API calls, no side effects. Verified on a CU401B_G + VitoCharge setup with two RoomControl devices present but no Zigbee sensors.

**For installations without RoomControl at all:** The enrichment loop simply finds no `roomControl` devices and skips entirely. Zero impact.

## `isSupported` handling

Features that only work with enrichment (CO2, room name, heating targets, etc.) now properly integrate with PyViCare's `@handleNotSupported` decorator. Without enrichment, `_getRoomControl()` raises `KeyError`, which the decorator converts to `PyViCareNotSupportedFeatureError`. This means HA's `is_supported()` check correctly reports these features as unsupported — no entity is created, no silent `None` values. Tested with `isSupported()` assertions and exception checks.

PyViCare/PyViCareDeviceConfig.py
```diff
@@ -113,6 +127,7 @@ def asAutoDetectDevice(self):
             (self.asRadiatorActuator, r"E3_RadiatorActuator", ["type:radiator"]),
             (self.asFloorHeating, r"Smart_zigbee_fht_main|E3_FloorHeatingCircuitDistributorBox", ["type:fhtMain"]),
             (self.asFloorHeatingChannel, r"Smart_zigbee_fht_channel", ["type:fhtChannel"]),
+            (self.asRoomControl, r"E3_RoomControl|Smart_RoomControl", ["type:virtual;smartRoomControl"]),
```
CFenner, line 130: Will this class only represent the virtual device that handles the room thermostat mapping or also in-room control real devices?

lackas:
Addressed review comments:
- Moved `_room_control` and `_room_id` to class-level defaults (no more `__init__`)
- Added type hints to all RoomSensor methods
- Fixed pylint `too-many-instance-attributes` on PyViCareDeviceConfig
- Rebased on updated #742

Re: RoomControl scope question - it only represents the virtual device (role `type:virtual;smartRoomControl`). It holds the room-to-sensor mapping and room-level data (temperature, humidity, CO2, schedules). The physical in-room controllers (Vitotrol 300E etc.) are separate zigbee devices with their own device type.

tests/test_TestForMissingProperties.py
```diff
@@ -300,6 +300,39 @@ def test_missingProperties(self):
             'fht.configuration.floorHeatingDamageProtectionThreshold',
             'fht.valve',
 
+            # RoomControl - room sensor data, used by enrichment (#743)
+            'rooms',
+            'rooms.0',
```
CFenner: Is this limited to 10 rooms?

PyViCare/PyViCareRoomSensor.py
CFenner, line 1: The room thermostat can be enhanced the same way no?

tests/test_TestForMissingProperties.py
lackas: No, it's not limited to 10 rooms. I just realized these ignore entries are actually unnecessary -- `find_feature_in_code` already replaces single digits with `{.*?}`, so `rooms.5.sensors.temperature` in test data matches `f"rooms.{room_id}.sensors.temperature"` in code. The test passes without them. I'll remove the entire block.

PyViCare/PyViCareDeviceConfig.py
lackas: Only the virtual device (role `type:virtual;smartRoomControl`). Physical in-room controllers like Vitotrol 300E are separate zigbee devices with their own device type and class.

PyViCare/PyViCareRoomSensor.py
lackas: Yes, any device that shows up as an actor in a RoomControl room could be enriched the same way. Right now we only do it for RoomSensor, but extending to RadiatorActuator or FloorHeating would be straightforward -- same `setRoomControlEnrichment` pattern. Happy to add that in a follow-up if there's demand.

CFenner: Ok, then let's go with this first.

CFenner (approved):

Issue #746, opened 2026-04-20
Cooling Mode not available


bekra00:
I need in HomeAssistant the cooling mode to use this for my Automations in KNX.
Therefore it is necessary to add following  sensor which is not yet available in PyViCare:
heating.circuits.N.operating.modes.cooling

lackas:
Hi @bekra00,

a summary of what we found while digging into this.

`heating.circuits.N.operating.modes.cooling` turned out to be a mode-selection/config flag, not a runtime state. On our own Vitocal 300-G (CU401B) I watched natural cooling run live (the buffer dropped from 42 to 19 °C) while `heating.circuits.1.operating.modes.cooling` stayed `active: false`, and the selected mode was the umbrella `dhwAndHeatingCooling`, so `getActiveMode()` never returns `cooling`.

Unfortunately, on our setup there is no clean "cooling is running now" flag in the API. Because it is natural cooling (no compressor), the compressor metrics stay 0 as well: `getCoefficientOfPerformanceCooling()` = 0 and `power.consumption.cooling` = 0. The only way we could tell cooling was active was the temperature signature, the circuit supply / system return / buffer dropping well below heating level (secondary supply went from ~42 °C to ~16 °C). Those thresholds are specific to each installation, so they do not generalize into a reliable getter, which is a pity.

For active cooling (compressor reversal, e.g. Vitocal 250-A) it should be easier: `getCoefficientOfPerformanceCooling()` > 0 or `CoolingCircuit.getReverseActive()` should indicate it.

To see whether a general indicator exists, could you (and anyone else here) help with a bit of data:

- Which device model, and active or natural cooling?
- A diagnostics dump (or the raw feature list) once with cooling clearly running and once without, so we can diff which feature actually flips.
- If you spot any feature on your device that reads true or nonzero only while cooling, please paste it here.

With a few examples across setups we can tell whether there is a common signal worth exposing, instead of a per-installation temperature heuristic.

PR #747, opened 2026-04-20
Added heating.circuits.N.operating.modes.cooling

bekra00:
Added heating.circuits.N.operating.modes.cooling
to the PyViCareHeating Device and setup testing


CFenner:
This can already be retrieved with `getActiveMode()`, not as a bool, but I think you get the idea. Do you still think we need a dedicated function for that?

bekra00:
By getActiveMode(), I receive dhwAndHeatingCooling or heating, but not if **in this moment** it operates in cooling. 

getActiveMode() -> heating.circuits.N.operating.modes.active
Shows current active **operating mode** on the device and provides command to change it.

getCoolingActive() -> heating.circuits.N.operating.modes.cooling
Shows whether the cooling operating mode **is active now**.

So it can be the case, that if the operation mode is set to dhwAndHeatingCooling, cooling is on or off. Therefore it does not provide the same information.

I want to use this value to open the zone valves.

CFenner:
Ok, then would it make sense to expose heating as well?

bekra00:
In general yes - not for my use case but it make sense

lackas:
`heating.circuits.N.operating.modes.cooling` is device-dependent. Where `cooling` is a directly selectable mode it flips `active: true`, but where the operating mode is an umbrella (`dhwAndHeatingCooling` / `heatingCooling`) and the pump decides heating vs cooling itself, `operating.modes.cooling.active` stays `false` even while cooling runs. I verified this on a Vitocal 300-G (CU401B) doing natural cooling: the buffer dropped from 42 to 19 °C while that flag stayed `false`, so `getCoolingActive()` would report the opposite of reality there.

So this works for some setups but not all. Might be worth naming it after what it actually reads (the configured cooling mode) rather than "active", and documenting the scope. Background in #746.

CFenner (approved):
👍

PR #750, opened 2026-04-28
feat: add all_devices property with unfiltered device list

lackas:
## Summary
- Add `all_devices` property that includes every device from the installation, regardless of `deviceType`
- Keep `devices` filtered to `SUPPORTED_DEVICE_TYPES` (no behavior change for existing consumers)
- Extract supported types as `SUPPORTED_DEVICE_TYPES` class constant

## Motivation
Devices with unknown `deviceType` (e.g. Vitoset Aqua with type `domesticHotWater`) are currently dropped by the filter in `__extract_devices()`. This means they don't appear in HA diagnostics, making it harder to debug and add support for new device types.

With `all_devices`, HA diagnostics can switch to showing all devices while entity setup continues using the safe filtered `devices` list.

## Follow-up
HA Core vicare integration needs a one-line change in diagnostics to use `all_devices` instead of `devices`. That PR will depend on a PyViCare release containing this change.

## Test plan
- All existing tests pass (no behavior change for `devices`)
- `all_devices` is a superset of `devices`

### DIFF
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -46,20 +46,24 @@ class PyViCare:
 
         data = installations['data']
         self.installations = Wrap(data)
-        self.devices = list(self.__extract_devices())
+        self.all_devices = list(self.__extract_all_devices())
+        self.devices = [d for d in self.all_devices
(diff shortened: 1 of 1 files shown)


CFenner (approved):
👍

PR #754, opened 2026-05-05
chore: add obtain_token_via_basic_auth_pkce migration helper

lackas:
Adds a public helper that exchanges stored username/password credentials for an OAuth2 token via Viessmann's PKCE auth-code flow with HTTP Basic auth. The authorization endpoint accepts Basic auth and returns the auth code in the redirect `Location` header, so no browser is involved.

## Motivation

The HA Core ViCare integration is migrating from username/password to OAuth2 with Application Credentials (home-assistant/core#165621). To avoid forcing every existing user to re-authenticate manually, a one-shot helper exchanges their stored credentials for a refresh token in the background.

Reviewer @joostlek pointed out that this is API code (Viessmann-specific protocol logic) and belongs in PyViCare, not in the integration. Moving it here.

## API

```python
from PyViCare.PyViCareOAuthManager import obtain_token_via_basic_auth_pkce

token = obtain_token_via_basic_auth_pkce(client_id, username, password)
# {} on any failure, otherwise dict with access_token, refresh_token, expires_at, ...
```

Failures (auth server unreachable, credentials rejected, token exchange failed) are logged at WARNING and return `{}`. Caller can detect the migration failure and prompt for reauth.

## Tests

Four new tests cover: success path, network failure during authorization, non-redirect response (bad credentials), and token exchange failure. All 737 PyViCare tests pass.

## Followup

Once this is merged and released, home-assistant/core#165621 will bump the PyViCare dependency and replace the integration-local copy with an import.

### DIFF
--- a/PyViCare/PyViCareOAuthManager.py
+++ b/PyViCare/PyViCareOAuthManager.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import pickle
+import warnings
 from contextlib import suppress
 from pickle import UnpicklingError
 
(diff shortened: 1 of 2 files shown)


CFenner (approved):

CFenner (approved):
👍

PR #761, opened 2026-05-09
feat: add WaterTreatment class for Vitoset Aqua

lackas:
**Type:** New feature

Adds read-only support for the Vitoset Aqua water treatment station.

The Vitoset Aqua exposes four functional areas in its IoT-scope feature dump:

- **Softener:** salt days remaining, low-salt alert threshold
- **Consumption:** current/max flow, daily / 7-day / total volume
- **Leak detection:** up to 5 sensors (status, leak, battery, RSSI, name, hardware/software versions) plus configurable flow alert
- **Shutoff valve:** position, motor state, holiday-mode flag

Autodetect matches device model `VitosetAqua*` (e.g. `VitosetAqua19D`, `VitosetAqua42D`) and the `type:waterTreatment` role.

**Setters intentionally deferred:** the IoT-scope dump contains no command definitions for this device, so write operations cannot currently be implemented. Will follow up if/when commands surface.

Test data was added separately in #756.

Tests: 743 pass, includes new `test_VitosetAqua.py` (82 lines) and 3 autodetect tests.

### DIFF
--- a/PyViCare/PyViCareWaterTreatment.py
+++ b/PyViCare/PyViCareWaterTreatment.py
@@ -1,0 +1,118 @@
+from typing import Any
+
+from PyViCare.PyViCareDevice import Device
+from PyViCare.PyViCareUtils import PyViCareNotSupportedFeatureError, handleNotSupported
+
+
+_LEAK_SENSOR_SLOTS = 5
+
+
+class WaterTreatment(Device):
+    """Viessmann Vitoset Aqua water treatment station.
+
+    Combines water softener, leak detection (up to 5 sensors), consumption metering
+    and main shutoff valve in one device.
+    """
+
+    # --- Softener ---
+
+    @handleNotSupported
+    def getSaltDaysRemaining(self) -> int:
+        return int(self.getProperty("water.softener.salt.level.days")["properties"]["remaining"]["value"])
+
+    @handleNotSupported
+    def getLowSaltAlertDays(self) -> int:
+        return int(self.getProperty("water.softener.configuration.lowSaltAlert")["properties"]["lowLevelAlertDays"]["value"])
+
+    # --- Consumption ---
+
+    @handleNotSupported
+    def getCurrentFlow(self) -> float:
+        return float(self.getProperty("water.consumption.flow.current")["properties"]["value"]["value"])
+
+    @handleNotSupported
+    def getMaxFlow(self) -> float:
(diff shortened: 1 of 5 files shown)


PyViCare/PyViCareWaterTreatment.py
```diff
@@ -0,0 +1,118 @@
...
+    def getMaxFlow(self) -> float:
+        return float(self.getProperty("water.consumption.flow.max")["properties"]["value"]["value"])
+
+    @handleNotSupported
+    def getCurrentDayConsumption(self) -> int:
+        return int(self.getProperty("water.consumption.summary")["properties"]["currentDay"]["value"])
+
+    @handleNotSupported
+    def getLastSevenDaysConsumption(self) -> int:
+        return int(self.getProperty("water.consumption.summary")["properties"]["lastSevenDays"]["value"])
```
CFenner: Can we omit the daily / weekly end points? In my eyes it's not really necessary for HA, which is the first-class user of the this library.

PyViCare/PyViCareWaterTreatment.py
```diff
@@ -0,0 +1,118 @@
...
+            if not base:
+                continue
+            id_props = self._readProperties(f"water.leakDetection.sensors.leakage.{slot}.id")
+            name_props = self._readProperties(f"water.leakDetection.sensors.leakage.{slot}.name")
+            battery_props = self._readProperties(f"water.leakDetection.sensors.leakage.{slot}.battery")
+            rssi_props = self._readProperties(f"water.leakDetection.sensors.leakage.{slot}.rssi")
+            hw_props = self._readProperties(f"water.leakDetection.sensors.leakage.{slot}.version.hardware")
+            sw_props = self._readProperties(f"water.leakDetection.sensors.leakage.{slot}.version.software")
+            sensors.append({
+                "slot": slot,
```
CFenner: These are not exposed as dedicated devices? What do you think of exposing them in a similar way than we do it with burners and compressors?

lackas:
Both points are addressed:
- Dropped the daily/weekly consumption endpoints (`64b8c5d`).
- Leak sensors are now exposed via a dedicated `LeakSensor` class, mirroring how burners and compressors are handled (`bb95a21`).

Ready for another look when you have a moment.

CFenner (approved):

PR #764, opened 2026-05-21
fix: catch PyViCareNotPaidForError in auto-detection and diagnostics

lackas:
Closes the integration-setup crash reported in home-assistant/core#167367.

## Problem

PR #737 introduced `PyViCareNotPaidForError` and made the cached service handle it gracefully, but two paths still propagate the exception unhandled:

1. **`_isHybridByFeatures()`** calls `fetch_all_features()` and catches only `KeyError/TypeError/AttributeError/OSError`. For GazBoiler and HeatPump devices the hybrid detection runs during `asAutoDetectDevice()`, so the exception crashes integration setup before any device is created. This matches the stacktrace reported by @drveni on home-assistant/core#167367 ending at `__handle_not_paid_for`.
2. **`dump_secure()` / `get_raw_json()`** have no exception handling at all, so the diagnostics endpoint crashes for the same users.

## Fix

- `_isHybridByFeatures()`: catch `PyViCareNotPaidForError` and fall back to non-hybrid. Downstream feature access still surfaces paid-only features as `PyViCareNotSupportedFeatureError` through the cached service.
- `dump_secure()`: catch `PyViCareNotPaidForError` and emit a placeholder dump with `device.dataUnavailableReason = "PACKAGE_NOT_PAID_FOR"` so diagnostics produce a usable file instead of crashing.

## Tests

- `test_autoDetect_NotPaidFor_keeps_original`: Vitocal auto-detects as `HeatPump` despite the API raising `PyViCareNotPaidForError`.
- `test_dump_secure_NotPaidFor_emits_placeholder`: diagnostics dump contains the `dataUnavailableReason` marker and an empty `data` list.

All 727 tests pass.

### DIFF
--- a/PyViCare/PyViCareDeviceConfig.py
+++ b/PyViCare/PyViCareDeviceConfig.py
@@ -16,6 +16,7 @@ from PyViCare.PyViCareRoomSensor import RoomSensor
 from PyViCare.PyViCareRepeater import Repeater
 from PyViCare.PyViCareElectricalEnergySystem import ElectricalEnergySystem
 from PyViCare.PyViCareGateway import Gateway
+from PyViCare.PyViCareUtils import PyViCareNotPaidForError
 from PyViCare.PyViCareVentilationDevice import VentilationDevice
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


CFenner (approved):
👍

PR #774, opened 2026-06-15
refactor: move accessor from service to device

lackas:
Splits the device-identity concern (installation/gateway/device IDs) from the HTTP/cache concern (OAuth, request building, caching). Service instances no longer hold an accessor — it lives on `Device` and `PyViCareDeviceConfig` and is passed at each call site.

Picked up from @CFenner's #628 on current master. The 8-month drift, especially #764's `PACKAGE_NOT_PAID_FOR` / `DeviceCommunicationError` / `InternalServerError` exception handling added to `ViCareCachedService`, made a fresh implementation cleaner than a rebase. Same design intent and API shape as #628 — closes #628.

## Motivation

This is prep work for a gateway-scoped service that can serve multiple devices from a single bulk fetch (`/features/installations/{id}/gateways/{serial}/features?includeDevicesFeatures=true`). Today's per-device `fetch_all_features()` is called N times per coordinator refresh; with a gateway service it could collapse to 1.

Measured on my installation: the bulk endpoint covers every `isEnabled=true` feature on every device (deactivated features are excluded by default, but PyViCare's existing `@handleNotSupported` / `isEnabled` checks already make missing and disabled features behaviourally equivalent). So a per-gateway architecture is functionally lossless for HA's use case — see also home-assistant/core#173776 where this came up.

## What changes

- `Device.__init__(accessor, service)` — stores `self.accessor`, threads it through `getProperty` / `setProperty`
- `PyViCareDeviceConfig.__init__(accessor, service, ...)` — `device_id` derived from `accessor.device_id`; all `as*()` constructors pass `self.accessor` through
- `ViCareService` and `ViCareCachedService` drop the `accessor` instance attribute; `getProperty` / `setProperty` / `fetch_all_features` / `buildGetPropertyUrl` / `reboot_gateway` take accessor as a parameter
- `get_available_burners(device)` (free function in `PyViCareHeatingDevice`) takes a device instead of a service — its 4 callers updated accordingly
- 21 sites in `PyViCareRoomControl` and 1 in `PyViCareHeatingDevice` that called `self.service.getProperty/setProperty` directly now go through the Device delegation (`self.getProperty/setProperty`), which was already the established pattern in the base class
- `ViCareCachedService` keeps the full defensive try/except for `PyViCareNotPaidForError`, `PyViCareDeviceCommunicationError`, `PyViCareInternalServerError` introduced in #764

## What stays the same

Public API surface used by consumers (HA Core, custom integrations):

- `device.getProperty(name)` / `device.setProperty(name, action, data)` — unchanged signatures
- `DeviceConfig.as*()` accessors — unchanged signatures, same return types
- `DeviceConfig.getConfig()` still returns the accessor

Code that constructs `ViCareService` / `ViCareCachedService` / `PyViCareDeviceConfig` directly will need the new signature. The `PyViCare` orchestrator itself handles the wiring.

## Verification

- All 729 tests pass (30 skipped, unchanged from master)
- `mypy PyViCare/` clean (the 3 missing-stub warnings are pre-existing env-only issues that CI handles)
- `ruff check PyViCare/ tests/` — all checks passed

## Credit

Originally proposed by @CFenner in #628 — this PR carries the same design forward on current master.

closes https://github.com/openviess/PyViCare/pull/628

### DIFF
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -32,10 +32,10 @@ class PyViCare:
     def initWithBrowserOAuth(self, client_id: str, token_file: str) -> None:
         self.initWithExternalOAuth(ViCareBrowserOAuthManager(client_id, token_file))
 
-    def __buildService(self, accessor, roles):
+    def __buildService(self, roles):
         if self.cacheDuration > 0:
-            return ViCareCachedService(self.oauth_manager, accessor, roles, self.cacheDuration)
-        return ViCareService(self.oauth_manager, accessor, roles)
+            return ViCareCachedService(self.oauth_manager, roles, self.cacheDuration)
+        return ViCareService(self.oauth_manager, roles)
 
     def __loadInstallations(self):
         installations = self.oauth_manager.get(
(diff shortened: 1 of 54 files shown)


lackas:
Rebased to drop the deprecation-checker commit — it's now a separate single-purpose PR #775 since it's orthogonal to this refactor. The `Format` step will stay red here until #775 lands, same as on master since #707.

CFenner (approved):
👍

PR #777, opened 2026-06-18
feat: add viaGateway mode for shared per-gateway service

lackas:
## Summary

Adds an opt-in service mode where ONE `ViCareCachedServiceViaGateway` instance serves all devices on a gateway from a single bulk API call, instead of N per-device calls per refresh.

```python
vicare = PyViCare()
vicare.loadViaGateway(True)   # default False; must precede initWith*
vicare.initWithCredentials(...)
```

Per-device entry points (`device.getProperty`, `DeviceConfig.as*`) are unchanged. Default behavior is unchanged.

## Why

The Viessmann API exposes a per-gateway bulk endpoint:
```
GET /features/installations/{id}/gateways/{serial}/features?includeDevicesFeatures=true
```
It returns all `isEnabled=true` features for every device on the gateway in one call. For overlapping features the property values + timestamps are byte-identical to per-device fetches.

For free-tier users (1450 API calls/day) this is meaningful headroom: a 4-device gateway drops from 4 calls per refresh to 1.

## Architecture

```
ViCareService                       (per-device, transport)
└── ViCareCachedService             (per-device, with cache; default)

ViCareServiceViaGateway             (per-gateway, transport)
└── ViCareCachedServiceViaGateway   (per-gateway, with cache; opt-in)
```

The shared service is wired in `PyViCare.__extract_all_devices`: one service per gateway in viaGateway mode, one per device otherwise. `setProperty` still uses the per-device URL in both modes (writes target one feature on one device; no bulk write endpoint exists).

Both `getProperty` and `fetch_all_features` are cached against the same payload. The `fetch_all_features` cache is the key optimization for HA Core's DataUpdateCoordinator pattern (per-device coordinators call `service.clear_cache + service.fetch_all_features` per refresh — without caching that path, N coordinators on a shared service would still trigger N bulk fetches per cycle).

Defensive error handling (`PACKAGE_NOT_PAID_FOR`, `DeviceCommunicationError`, `InternalServerError`) mirrors `ViCareCachedService`.

## Precursor refactor

Commit 1 moves role-based decisions (`hasRoles`, `isGateway`, `isLegacyDevice`, `isE3Device`, `get_heat_curve_formular`) from Service to Device/DeviceConfig. In shared-service mode the service cannot answer per-device role questions because its roles list would be whichever device was processed first. Service-side `hasRoles`/`_isGateway` are kept for backwards compatibility (still used by the per-device service for URL shape selection).

Public API unchanged. Test mocks updated to pass roles via the `PyViCareDeviceConfig` constructor instead of patching `service.hasRoles`.

## Credit

Carries the design intent of #626 by @CFenner forward. The bulk response test fixtures (heatbox1/heatbox2/tcu1) are lifted from that draft. Implementation is fresh on current master because #774 made the previous sketch incompatible (accessor lives on Device now, service is stateless), and the original draft did not actually share the service instance across devices.

## Tests

20 new tests covering URL building, per-device filtering of the bulk response, shared-cache behavior across devices, TTL respect, `setProperty` cache invalidation, stale-cache fallback on transient errors, end-to-end PyViCare wiring with mocked OAuth.

749 → 750 total tests passing, ruff clean, pylint 10/10 on changed modules.

Refs: #626, #774, home-assistant/core#173776

### DIFF
--- a/PyViCare/PyViCare.py
+++ b/PyViCare/PyViCare.py
@@ -4,9 +4,11 @@ from datetime import datetime
 from PyViCare.PyViCareAbstractOAuthManager import AbstractViCareOAuthManager
 from PyViCare.PyViCareBrowserOAuthManager import ViCareBrowserOAuthManager
 from PyViCare.PyViCareCachedService import ViCareCachedService
+from PyViCare.PyViCareCachedServiceViaGateway import ViCareCachedServiceViaGateway
 from PyViCare.PyViCareDeviceConfig import PyViCareDeviceConfig
 from PyViCare.PyViCareOAuthManager import ViCareOAuthManager
 from PyViCare.PyViCareService import ViCareDeviceAccessor, ViCareService
+from PyViCare.PyViCareServiceViaGateway import ViCareServiceViaGateway
 from PyViCare.PyViCareUtils import PyViCareInvalidDataError
 
 logger = logging.getLogger(__name__)
@@ -17,9 +19,21 @@ class PyViCare:
     """"Viessmann ViCare API Python tools"""
     def __init__(self) -> None:
         self.cacheDuration = 60
+        self.viaGateway = False
 
     def setCacheDuration(self, cache_duration):
         self.cacheDuration = int(cache_duration)
+
+    def loadViaGateway(self, via_gateway: bool = True) -> None:
+        """Opt in to per-gateway bulk fetching.
+
+        When enabled, ONE service instance per gateway serves all devices on
+        that gateway via the bulk endpoint
+        `/features/installations/{id}/gateways/{serial}/features?includeDevicesFeatures=true`,
+        instead of one HTTP call per device per refresh. Must be called
+        before initWith* (services are wired during __loadInstallations).
+        """
+        self.viaGateway = via_gateway
 
     def initWithCredentials(self, username: str, password: str, client_id: str, token_file: str):
         self.initWithExternalOAuth(ViCareOAuthManager(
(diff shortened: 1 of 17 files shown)


PyViCare/PyViCare.py
```diff
@@ -17,10 +19,22 @@ class PyViCare:
     """"Viessmann ViCare API Python tools"""
     def __init__(self) -> None:
         self.cacheDuration = 60
+        self.viaGateway = False
```
CFenner: Can be moved out of `init`.

PyViCare/PyViCareServiceViaGateway.py
```diff
@@ -0,0 +1,38 @@
...
+
+from PyViCare.PyViCareAbstractOAuthManager import AbstractViCareOAuthManager
+from PyViCare.PyViCareService import (ViCareDeviceAccessor, ViCareService,
+                                      readFeature)
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+
+def filter_features_for_device(entities, device_id: str):
```
CFenner: define type of `entities`

PyViCare/PyViCareCachedServiceViaGateway.py
```diff
@@ -0,0 +1,92 @@
...
+                        return self.__cache
+                    raise
+
+                if "data" not in data:
+                    logger.error("Missing 'data' property when fetching data.")
+                    raise PyViCareInvalidDataError(data)
+                self.__cache = data
+            return self.__cache
+
+    def is_cache_invalid(self) -> bool:
```
CFenner: Can we have a common `cachedService` class to not duplicate the caching code?

PyViCare/PyViCareDeviceConfig.py
```diff
@@ -101,8 +101,17 @@ def getDeviceType(self):
     def getRoles(self):
         return self.roles
 
+    def hasRoles(self, requested_roles):
+        return hasRoles(requested_roles, self.roles)
+
     def isGateway(self):
-        return self.service._isGateway()  # pylint: disable=protected-access
```
CFenner, line 105: Can we then remove `_isGateway`?

CFenner (approved):
👍

PR #781, opened 2026-06-21
feat: expose ventilation bypass properties

Ingrater:
Added support for the following three properties on ventilation devices:

- ventilation.bypass.position
- ventilation.bypass.operating.modes.open
- ventilation.bypass.operating.modes.automatic

The bypass position indicates how much of the air is being fed into the heat recovery mechanism in the ventilation device. Where 0 = all of the air, 100 = none of it. E.g. how much air "bypasses" the heat recovery. In summer this is a critical value to understand if the ventilation system is optimally reducing the temperature in the house.

The open mode, is essentially the "manual" mode of bypass configuration. The automatic mode on the other side means the system is automatically controlling the bypass. Useful to see how the ventilation device was configured.

It was not required to add new test data as a dump of the VitoairFS300 which I own is already checked in.

As a next step I plan on extending the ViCare Home Assistant integration to expose these values as well.

### DIFF
--- a/PyViCare/PyViCareVentilationDevice.py
+++ b/PyViCare/PyViCareVentilationDevice.py
@@ -354,3 +354,15 @@ class VentilationDevice(Device):
     @handleNotSupported
     def getConfiguredLevelFourVolumeFlow(self) -> int:
         return int(self.getProperty("ventilation.levels.levelFour")["properties"]["volumeFlow"]["value"])
+
+    @handleNotSupported
+    def getBypassPosition(self) -> int:
+        return int(self.getProperty("ventilation.bypass.position")["properties"]["value"]["value"])
+
+    @handleNotSupported
+    def getBypassOperatingModesOpen(self) -> bool:
+        return bool(self.getProperty("ventilation.bypass.operating.modes.open")["properties"]["active"]["value"])
+
+    @handleNotSupported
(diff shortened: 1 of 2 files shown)


PyViCare/PyViCareVentilationDevice.py
```diff
@@ -354,3 +354,15 @@ def getConfiguredLevelThreeVolumeFlow(self) -> int:
     @handleNotSupported
     def getConfiguredLevelFourVolumeFlow(self) -> int:
         return int(self.getProperty("ventilation.levels.levelFour")["properties"]["volumeFlow"]["value"])
+
+    @handleNotSupported
+    def getBypassPosition(self) -> int:
+        return int(self.getProperty("ventilation.bypass.position")["properties"]["value"]["value"])
+
+    @handleNotSupported
+    def getBypassOperatingModesOpen(self) -> bool:
```
CFenner, line 363: Can you rather expose the `*.active.value` attribute as we have it for other modes. Also you could expose the `*.active.state` and `*.active.level` as well.

CFenner (commented):
Thanks for the PR! ❤️ 

Please remove exposed attributes from https://github.com/openviess/PyViCare/blob/9cefbfe616e52b7f0eca79a2df092c14d9cdc3e0/tests/test_TestForMissingProperties.py#L260

PR #791, opened 2026-07-07
feat: enable getBoilerTemperature for Oil and PelletBoilers

### LINKED ISSUE #477: commonSupply is present in almost all units, but only exposed for GazBoiler
Grepping the dumps in tests/response for `commonSupply` as exposed by [GazBoiler.getBoilerCommonSupplyTemperature](https://github.com/openviess/PyViCare/blob/4e716c71f9ae3a4953852e5a3857711cd5947024/PyViCare/PyViCareGazBoiler.py#L107), this property seems to be present in almost all units. 

At least for the Vitocal250, I can confirm, that it is present in the dump and also in my device. Furthermore the number seems to be shown in the German version of the App as "Vorlauftemperatur Wärmeerzeuger", right next to `getReturnTemperature()` ("Rücklauftemperatur Wärmeerzeuger").

Workaround to get this number - retrieve the service from any PyViCare object and query it directly

```
commonSupply = heatpump.service.getProperty("heating.boiler.sensors.temperature.commonSupply")["properties"]["value"]["value"]
```

lackas:
Closes #477 (the remaining part).

`getBoilerCommonSupplyTemperature` was already moved to `HeatingDevice`. This does the same for `getBoilerTemperature` (`heating.boiler.sensors.temperature.main`), which is present on oil (Vitoladens) and pellets (Vitoligno/Ecotronic) boilers too but was only reachable on `GazBoiler`. `OilBoiler` and `PelletsBoiler` extend `HeatingDevice` directly, so they could not call it before.

Heat pumps do not expose `...temperature.main` (they use `commonSupply`), so this does not affect Vitocal devices. Test added on the Ecotronic (PelletsBoiler) fixture; GazBoiler keeps the method by inheritance and its tests stay green.

### DIFF
--- a/PyViCare/PyViCareGazBoiler.py
+++ b/PyViCare/PyViCareGazBoiler.py
@@ -92,10 +92,6 @@ class GazBoiler(HeatingDevice):
         return self.getProperty("heating.gas.consumption.dhw")["properties"]["year"]["value"][0]
 
     @handleNotSupported
-    def getBoilerTemperature(self):
-        return self.getProperty("heating.boiler.sensors.temperature.main")["properties"]["value"]["value"]
-
-    @handleNotSupported
     def getBoilerTargetTemperature(self):
         return self.getProperty("heating.boiler.temperature")["properties"]["value"]["value"]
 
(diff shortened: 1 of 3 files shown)


CFenner (approved):
👍

